PR #29, opened 2022-03-15
Fix event scope.

Kerilk:
Despite: https://spec.oneapi.io/level-zero/latest/core/PROG.html#kernel-timestamp-events


TApplencourt:
Sadly LGTM :(

PR #51, opened 2023-01-11
Debug

Kerilk:


ze/tracer_ze_helpers.include.c
```diff
@@ -1,8 +1,24 @@
+#ifdef THAPI_DEBUG
+#define TAHPI_LOG stderr
+#define THAPI_DBGLOG(fmt, ...) \
+ do { \
+   fprintf(TAHPI_LOG, "THAPI(%s:%d): " fmt "\n", __func__, __LINE__, __VA_ARGS__); \
+ } while(0)
+#define THAPI_DBGLOG_NO_ARGS(fmt) \
```
TApplencourt, line 7: Look like we can use https://stackoverflow.com/a/59044559/7674852, to remove the `NO_ARGS`

ze/tracer_ze_helpers.include.c
```diff
@@ -274,9 +308,13 @@ static inline void _register_ze_event(
   if (o_h) {
     cl_data = (struct _ze_command_list_obj_data *)(o_h->obj_data);
     _ze_event->context = cl_data->context;
-  }
+    if (cl_data->flags & _ZE_IMMEDIATE)
+      _ze_event->flags |= _ZE_IMMEDIATE_CMD;
+  } else
+    THAPI_DBGLOG("Could not get command list associated to event: %p", event);
```
TApplencourt, line 314: Just by convension should will not always put `{}`arround macro?

ze/tracer_ze_helpers.include.c
```diff
@@ -325,27 +371,60 @@ static struct _ze_event_h * _get_profiling_event(
...
+#ifdef THAPI_DEBUG
+  struct _ze_obj_h *o_h = NULL;
+  FIND_ZE_OBJ(&event, o_h);
+  if (o_h) {
+    THAPI_DBGLOG("Event already registered: %p", event);
+    return;
+  }
+
+  intptr_t mem = (intptr_t)calloc(1, sizeof(struct _ze_obj_h));
+
```
TApplencourt, line 384: Can remove this space

ze/tracer_ze_helpers.include.c
```diff
@@ -325,27 +371,60 @@ static struct _ze_event_h * _get_profiling_event(
...
+  struct _ze_obj_h *o_h = NULL;
+  FIND_AND_DEL_ZE_OBJ(&event, o_h);
+  if (!o_h) {
+    THAPI_DBGLOG("Could not find event: %p", event);
+  }
+#endif
+
   FIND_AND_DEL_ZE_EVENT(&event, ze_event);
-  if (!ze_event)
+  if (!ze_event) {
```
TApplencourt, line 412: Should we print a warning?

ze/tracer_ze_helpers.include.c
Kerilk: `__VA_OPT__ (c++2a)`
It would not work in C mode with our current flags.

ze/tracer_ze_helpers.include.c
Kerilk: the macros are protected by `do { } while(0)` constructs for this specific reason.

ze/tracer_ze_helpers.include.c
Kerilk, line 384:
```suggestion
```

Kerilk: Done, thanks.

ze/tracer_ze_helpers.include.c
Kerilk: It is not really an error to destroy an event that was already profiled or reset or not used. The vent not being in the event to trace pool just means we should not ask for profiling infos.

This is different in the reset case where if the user resets the event, it should have triggered before and we should have profiled it (It may still trigger on user events that are not used for profiling, but forhost synchronization)

ze/tracer_ze_helpers.include.c
Kerilk: Yes, it is an extension, so our strict flags will trigger. Let's revisit once we move to later C standards.

Kerilk:
Good for me!

PR #76, opened 2023-04-25
Split libxprof into libxtimeline and libxtally

DonAurelio:
We split libxprof into two different components (btx_tally.c and btx_timeline.c). This is the preamble to migrate such components into metababel.


xprof/btx_tally.c
```diff
@@ -0,0 +1,17 @@
+#include <babeltrace2/babeltrace.h>
+#include "tally.h"
```
TApplencourt, line 2: Maybe rename it to `btx_tally` then?

xprof/btx_timeline.c
```diff
@@ -1,21 +1,10 @@
 #include <babeltrace2/babeltrace.h>
-#include "tally.h"
 #include "timeline.h"
```
TApplencourt, line 2: btx_timeline?

xprof/btx_tally.c
DonAurelio: Makes sense, since we have babeltrace code spread on these files. I understand proper babeltrace stuff should be btx_.....

TApplencourt:
LGTM!

TApplencourt (approved):

Kerilk (approved):
LGTM.

PR #94, opened 2023-06-21
Remove usage of ctf_array_text in favor of ctf_sequence_text.

Kerilk:


utils/meta_parameters.rb
```diff
@@ -53,7 +53,7 @@ def check_for_null(expr, incl = true)
     end
   end
 
-  def sanitize_expression(expr, checks = check_for_null(expr, false), default = 0)
```
TApplencourt, line 56: what? You can pass as arguments to a function used a default parameter another parameter.  mind blow

Kerilk: I know right!!!

TApplencourt (approved):
I trust than you installed it and run it :p because we have no test for this part :D

PR #98, opened 2023-06-23
We should send Begin and End messages

TApplencourt:
Kind of ugly, but the clean way is done by metababel.
This component is really horrible and should be replaced soon...


Kerilk (approved):
Looks good.

Issue #101, opened 2023-06-27
PR #98 introduced a bug in iprof


DonAurelio:
Seems PR #98 introduced a bug in iprof. 

```bash
../THAPI/build/ici/bin/iprof ./a.out 
Trace location: /home/avivasmeza/lttng-traces/iprof-20230627-135841

/soft/packaging/spack-builds/linux-opensuse_leap15-x86_64/gcc-10.2.0/ruby-2.7.1-g5tqnvcmjqmx7k5xj6gbs2yhwbkjeswc/lib/ruby/gems/2.7.0/gems/babeltrace2-0.1.1/lib/babeltrace2/graph/graph.rb:449: [BUG] Segmentation fault at 0x0000000000000000
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0004 p:---- s:0035 e:000034 CFUNC  :bt_graph_run
c:0003 p:0033 s:0030 e:000029 METHOD /soft/packaging/spack-builds/linux-opensuse_leap15-x86_64/gcc-10.2.0/ruby-2.7.1-g5tqnvcmjqmx7k5xj6gbs2yhwbkjeswc/lib/ruby/gems/
c:0002 p:0904 s:0025 E:001558 EVAL   /home/avivasmeza/dev/THAPI/build/ici/bin/babeltrace_thapi:212 [FINISH]
c:0001 p:0000 s:0003 E:000c40 (none) [FINISH]
...
```

Minimal reproducible example:
```c++
// $ module use /soft/modulefiles
// $ module load iprof intel/oneapi compiler/latest intel_compute_runtime/

// $ icpx -fiopenmp -fopenmp-targets=spir64 foo.cpp
// $ iprof ./a.out

#include <iostream>
#include <cstdlib>
#include <algorithm>
#include <limits>
#include <vector>
bool almost_equal(float x, float y, int ulp) {
  return std::abs(x-y) <= std::numeric_limits<float>::epsilon() * std::abs(x+y) * ulp || std::abs(x-y) < std::numeric_limits<float>::min();
}
void test_target_teams_distribute_parallel_for_simd() {
  const int N0 { 32768 };
  const int size = N0;
  std::vector<float> dst(size), src(size);
  std::generate(src.begin(), src.end(), std::rand);
  float *pS { src.data() };
  float *pD { dst.data() };
[...]

PR #108, opened 2023-07-07
Tally filter

TApplencourt:
This introduces the splitting of tally into a filter and and a sink.

- Needed to add a suffix to the makefile rule to distinguish between sink and filter.
- Duplicated some code in the model because handling multiple yaml-model in the makefile was painful. We can fix that later. (not sure how...)

- The babeltrace_thapi is definitely ugly, we should re-write it with some better abstraction. But didn't want to pollute the PR with that.


TApplencourt:
@DonAurelio  and @Kerilk  ready for you review guys.

xprof/tally.cpp
```diff
@@ -213,14 +214,17 @@ static void ust_thapi_metadata_usr_callback(void *btx_handle, void *usr_data,
...
+
 void btx_register_usr_callbacks(void *btx_handle) {
   btx_register_callbacks_initialize_usr_data(btx_handle, &btx_initialize_usr_data);
   btx_register_callbacks_read_params(btx_handle, &btx_read_params);
   btx_register_callbacks_finalize_usr_data(btx_handle, &btx_finalize_usr_data);
 
-  btx_register_callbacks_lttng_host(btx_handle, &host_usr_callback);
-  btx_register_callbacks_lttng_device(btx_handle, &device_usr_callback);
-  btx_register_callbacks_lttng_traffic(btx_handle, &traffic_usr_callback);
   btx_register_callbacks_lttng_device_name(btx_handle, &device_name_usr_callback);
```
Kerilk, line 223: Why is this one called `lttng_device_name`?

TApplencourt: Stupid Historical reason. Will be fixed, when we will migrate all the interval to metababel

Kerilk: Sounds good!

Kerilk (dismissed):
Looks good here!

xprof/tally.hpp
```diff
@@ -425,43 +379,18 @@ auto get_uniq_tally(Map<std::tuple<K...>, V> &input) {
...
+  for (auto const &[_, t] : m)
+    tot += t;
+  m.push_back( { "Total", tot } );
 
-  if (tot.error == tot.count)
-    tot.duration_ratio = NAN;
-
-  m.push_back(std::make_pair(std::string("Total"), tot));
+  for (auto &[_, t] : m)
+    t.compute_duration_ratio(tot);
```
DonAurelio, line 390:
```suggestion
    keyval.second.compute_duration_ratio(tot);
```

DonAurelio:
In this branch, we are still having this https://github.com/argonne-lcf/THAPI/issues/101 issue, seems it was not merged properly in this branch

xprof/tally.cpp
```diff
@@ -159,49 +159,50 @@ void btx_finalize_usr_data(void *btx_handle, void *usr_data) {
...
+                                uint64_t min, uint64_t max, uint64_t total, uint64_t count,
+                                uint64_t backend) {
+
   tally_dispatch_t *data = (tally_dispatch_t *)usr_data;
 
-  TallyCoreByte a{(uint64_t)size, false};
   const int level = data->backend_level[backend];
   data->traffic_backend_name[level].insert(backend_name[backend]);
-  data->traffic[level][hpt_function_name_t(hostname, vpid, vtid, name)] += a;
+  data->traffic[level][ {hostname, vpid, vtid, name} ] +=  {total, 0, count, min, max};;
```
DonAurelio, line 198: * Are `total` and `duration` (`dur`) the same? 
* Can we consider the use of `dur` in place of `total` in the code and the model.yaml to be consistent?

xprof/btx_timeline_params.yaml
```diff
@@ -1,2 +1,4 @@
 :type: map
-:entries: []
+:entries:
```
DonAurelio, line 2: I do not think this feature should be introduced on this PR

DonAurelio, line 2: I do not think this feature should be introduced on this PR, since we are doing nothing with this param at all.

xprof/btx_aggreg_params.yaml
DonAurelio, line 1: Why does aggregation need these params?

xprof/btx_aggreg_model.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - :name: tally_traffic
+        :payload_field_class:
+          :type: structure
+          :members:
+            - :name: backend
+              :field_class:
+                :type: integer_unsigned
+                :field_value_range: 64
+                :cast_type: uint64_t
+  - :name: interval
```
DonAurelio, line 86: Why do we need this stream class on this file?

xprof/btx_aggreg_model.yaml
```diff
@@ -0,0 +1,135 @@
+:stream_classes:
+  - :name: interval_tally
```
DonAurelio, line 2: maybe we can call this stream class `aggreg`?

xprof/btx_aggreg_model.yaml
```diff
@@ -0,0 +1,135 @@
...
+            :type: integer_unsigned
+            :field_value_range: 64
+            :cast_type: uint64_t
+        - :name: count
+          :field_class:
+            :type: integer_unsigned
+            :field_value_range: 64
+            :cast_type: uint64_t
+    :event_classes:
+      - :name: tally_host
```
DonAurelio, line 45: Maybe `aggreg:host` to be consistent?

xprof/btx_timeline_params.yaml
TApplencourt: I doesn't work, we need a new version of metababel (see some PR on metababel)

xprof/btx_aggreg_params.yaml
TApplencourt: I don't, It's just to avoid a metababel bug for now.

xprof/btx_aggreg_model.yaml
TApplencourt: Require to be able to read the lttng::device name

xprof/btx_aggreg_model.yaml
TApplencourt: Good point!

xprof/btx_aggreg_model.yaml
TApplencourt: Good point

xprof/tally.cpp
TApplencourt: Hum true, but we cannot really change the model and we will need to change in the `interval.c` too. Let me think

xprof/tally.hpp
```diff
@@ -425,43 +379,18 @@ auto get_uniq_tally(Map<std::tuple<K...>, V> &input) {
...
-
-  // Create the final Tally
   TC tot{};
-  for (auto const &keyval : m)
-    tot += keyval.second;
-  // Finalize (compute ratio)
-  for (auto &keyval : m)
-    keyval.second.finalize(tot);
+  for (auto const &[_, t] : m)
+    tot += t;
```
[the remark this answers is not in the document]
TApplencourt: Maybe then?
```
  for (auto &[_, t] : m)
    t.finalize(tot);
    tot += t;
```

xprof/tally.hpp
TApplencourt: Why do you prefer `keyval.second` ?

xprof/tally.hpp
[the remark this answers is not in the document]
DonAurelio: I agree on the use of `_`, but the computer don't, it says `_` is unused. So no other way to prevent the warning

xprof/tally.hpp
DonAurelio: Good question, I like your approach with the dummy variable `_`, but we still have the compiler indicating this variable is unused.

xprof/tally.cpp
DonAurelio: Oh, how the change from `total` to `dur` is related to `interval.c`?

xprof/btx_timeline_params.yaml
DonAurelio: I will take a look on that :)

xprof/tally.hpp
DonAurelio: * g++ (SUSE Linux) 7.5.0
* gcc (SUSE Linux) 7.5.0

xprof/tally.hpp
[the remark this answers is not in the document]
TApplencourt: Fixed with ef7f5d5

xprof/tally.hpp
TApplencourt: Fixed with ef7f5d5

xprof/tally.cpp
TApplencourt: I mean we have `duration` for the interval and `total` for the interval. Doesn't that make sense?

xprof/btx_aggreg_model.yaml
TApplencourt: Exacly. We can refactor everything  cleanly when we move the interval to metababel

xprof/btx_aggreg_model.yaml
TApplencourt: Fixed by 6e21598

xprof/tally.cpp
DonAurelio: You are right that makes sense.

DonAurelio:
We are good to go now

DonAurelio (approved):

Kerilk (approved):

PR #111, opened 2023-07-11
Fix pipefail

TApplencourt:
So because of pipefail the error of `broken pipe` (which can appear when doing something like `foo | head -1`)  will make all the programs stop silently.

This fix this by using `|| true` 

(Kind of important, as `iprof -r` is currently broken :( )


Kerilk (approved):
Looks good to me.

PR #123, opened 2023-08-03
Added missing cudart APIs.

Kerilk:
Still missing GL and EGL APIs.

### DIFF
--- a/cuda/extract_cudart.rb
+++ b/cuda/extract_cudart.rb
@@ -1,13 +1,52 @@
 require_relative 'extract_base.rb'
 
 begin
+  preprocessed_sources_libc = $cpp.preprocess(<<EOF).gsub(/^#.*?$/, '')
+#include <stdint.h>
+EOF
+rescue
+  C::Preprocessor.command = "gcc -E"
+  preprocessed_sources_libc = $cpp.preprocess(<<EOF).gsub(/^#.*?$/, '')
+#include <stdint.h>
+EOF
+end
+
+$parser.parse(preprocessed_sources_libc)
+
+begin
   preprocessed_sources_cudart_api = $cpp.preprocess(<<EOF).gsub(/^#.*?$/, '')
 #include <cuda_runtime_api.h>
+#include <__cudart.h>
+typedef int32_t VdpStatus;
+typedef uint32_t VdpFuncId;
+typedef uint32_t VdpDevice;
+typedef VdpStatus  VdpGetProcAddress(
+     VdpDevice device,
+     VdpFuncId function_id,
+     void * *  function_pointer
+);
+typedef uint32_t VdpVideoSurface;
+typedef uint32_t VdpOutputSurface;
+#include <cuda_vdpau_interop.h>
+#include <cuda_profiler_api.h>
 EOF
 rescue
   C::Preprocessor.command = "gcc -E"
   preprocessed_sources_cudart_api = $cpp.preprocess(<<EOF).gsub(/^#.*?$/, '')
 #include <cuda_runtime_api.h>
+#include <__cudart.h>
+typedef int32_t VdpStatus;
+typedef uint32_t VdpFuncId;
+typedef uint32_t VdpDevice;
+typedef VdpStatus  VdpGetProcAddress(
+     VdpDevice device,
+     VdpFuncId function_id,
+     void * *  function_pointer
+);
+typedef uint32_t VdpVideoSurface;
+typedef uint32_t VdpOutputSurface;
+#include <cuda_vdpau_interop.h>
+#include <cuda_profiler_api.h>
 EOF
 end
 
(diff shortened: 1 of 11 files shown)


Kerilk:
@TApplencourt I think you can try again

TApplencourt:
Pass, put doesn't print any `cuda` call.
```
tapplencourt@gene-lxc:~/THAPI/build$ nvcc --cudart shared ~/test2.cu
tapplencourt@gene-lxc:~/THAPI/build$ cat ~/test2.cu

#include <stdio.h>

const int N = 16;
const int blocksize = 16;

__global__
void hello(char *a, int *b)
{
	a[threadIdx.x] += b[threadIdx.x];
}

int main()
{
	char a[N] = "Hello \0\0\0\0\0\0";
	int b[N] = {15, 10, 6, 0, -11, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};

	char *ad;
	int *bd;
	const int csize = N*sizeof(char);
	const int isize = N*sizeof(int);

	printf("%s BEFORE ", a);

	cudaMalloc( (void**)&ad, csize );
	cudaMalloc( (void**)&bd, isize );
	cudaMemcpy( ad, a, csize, cudaMemcpyHostToDevice );
	cudaMemcpy( bd, b, isize, cudaMemcpyHostToDevice );

	dim3 dimBlock( blocksize, 1 );
	dim3 dimGrid( 1, 1 );
	hello<<<dimGrid, dimBlock>>>(ad, bd);
	cudaMemcpy( a, ad, csize, cudaMemcpyDeviceToHost );
	cudaFree( ad );
	cudaFree( bd );

	printf("%s AFTER \n", a);
	return EXIT_SUCCESS;
}
tapplencourt@gene-lxc:~/THAPI/build$ ldd ./a.out
	linux-vdso.so.1 (0x00007f0a20ace000)
	libcudart.so.12 => /usr/local/cuda-12/targets/x86_64-linux/lib/libcudart.so.12 (0x00007f0a20800000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0a20400000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0a20ad0000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f0a207fb000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f0a207f6000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f0a207f1000)
tapplencourt@gene-lxc:~/THAPI/build$ ./a.out
Hello  BEFORE World! AFTER
tapplencourt@gene-lxc:~/THAPI/build$ ./ici/bin/tracer_cuda.sh --cudart ./a.out
Session thapi-cuda-session created.
Traces will be output to /home/tapplencourt/lttng-traces/thapi-cuda-session-20230803-152402
ust channel blocking-channel enabled for session thapi-cuda-session
ust context vpid added to channel blocking-channel
ust context vtid added to channel blocking-channel
ust event lttng_ust_cuda:* created in channel blocking-channel
ust event lttng_ust_cudart:* created in channel blocking-channel
Tracing started for session thapi-cuda-session
Failure: could not load cudart library!
Waiting for data availability
Tracing stopped for session thapi-cuda-session
Destroying session thapi-cuda-session..
Session thapi-cuda-session destroyed
tapplencourt@gene-lxc:~/THAPI/build$ babeltrace2 /home/tapplencourt/lttng-traces/thapi-cuda-session-20230803-152402
```

Kerilk:
Yeah seems it is not in the LD_LIBRARY_PATH

cuda/extract_cudart.rb
```diff
@@ -1,13 +1,52 @@
...
+#include <stdint.h>
+EOF
+rescue
+  C::Preprocessor.command = "gcc -E"
+  preprocessed_sources_libc = $cpp.preprocess(<<EOF).gsub(/^#.*?$/, '')
+#include <stdint.h>
+EOF
+end
+
+$parser.parse(preprocessed_sources_libc)
```
TApplencourt, line 14: Maybe:

```suggestion
begin
  preprocessed_sources_libc = $cpp.preprocess
#include <stdint.h>
EOF
rescue
  C::Preprocessor.command = "gcc -E"
  preprocessed_sources_libc = $cpp.preprocess
#include <stdint.h>
EOF
end

$parser.parse(preprocessed_sources_libc.).gsub(/^#.*?$/, ''))
```

Just to make clure that the gsub is not the problem, and the same in both case

cuda/extract_cudart.rb
```diff
@@ -1,13 +1,52 @@
...
+     VdpFuncId function_id,
+     void * *  function_pointer
+);
+typedef uint32_t VdpVideoSurface;
+typedef uint32_t VdpOutputSurface;
+#include <cuda_vdpau_interop.h>
+#include <cuda_profiler_api.h>
 EOF
 rescue
   C::Preprocessor.command = "gcc -E"
```
TApplencourt, line 34: We may want to create a function to to see

try if else add `gcc -E`, and then gsub

Kerilk: Yeah, but that would need to be in another PR since this pattern is used everywhere in the code. Maybe open an issue for this one?

TApplencourt (approved):

PR #136, opened 2023-08-25
Fix -Werror=maybe-uninitialized

Kerilk:
Before a recent c++ compiler complains (it did on my Ubuntu 23.04).


TApplencourt:
Stupid compiler with poor static analysis :(

TApplencourt (approved):

PR #137, opened 2023-08-25
Cleanup of babeltrace and lttng power/frequency code.

Kerilk:
Add awareness of hostname to the timeline.

### DIFF
--- a/xprof/timeline.cpp
+++ b/xprof/timeline.cpp
@@ -15,13 +15,6 @@ static perfetto_uuid_t gen_perfetto_uuid() {
   static std::atomic<perfetto_uuid_t> uuid{1};
   return uuid++;
 }
-bool pushIfNotExists(std::vector<uint64_t>& vec, uint64_t value) {
-    if (std::find(vec.begin(), vec.end(), value) == vec.end()) {
-        vec.push_back(value);
-        return true;
-    }
-    return false;
-}
 
 static void add_event_begin(struct timeline_dispatch *dispatch, perfetto_uuid_t uuid,
                             timestamp_t begin, std::string name) {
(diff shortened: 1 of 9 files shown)


xprof/timeline.cpp
```diff
@@ -44,6 +37,92 @@ static void add_event_end(struct timeline_dispatch *dispatch, perfetto_uuid_t uu
...
+  auto &potential_uuid = r.first->second;
+  // Exists
+  if (!r.second)
+    return potential_uuid;
+
+  perfetto_uuid_t hp_uuid = 0;
+  auto r2 = parent_tracks.insert({{hostname, process_id}, hp_uuid});
+  auto &potential_uuid2 = r2.first->second;
+  // Didn't exist
+  if (r.second) {
```
TApplencourt, line 54:
```suggestion
  if (r2.second) {
```

(may want to use scope)

Kerilk: Thanks!

xprof/timeline.cpp
```diff
@@ -44,6 +37,92 @@ static void add_event_end(struct timeline_dispatch *dispatch, perfetto_uuid_t uu
...
+  perfetto_uuid_t hp_uuid = 0;
+  auto r2 = parent_tracks.insert({{hostname, process_id}, hp_uuid});
+  auto &potential_uuid2 = r2.first->second;
+  // Didn't exist
+  if (r2.second) {
+    hp_uuid = gen_perfetto_uuid();
+    potential_uuid2 = hp_uuid;
+    // Create parent track
+    auto *packet = dispatch->trace.add_packet();
+    packet->set_trusted_packet_sequence_id(10000);
```
TApplencourt, line 59: Maybe use macro, better than magic number
(when doing the PR master)

PR #140, opened 2023-08-29
Add stub implementation of CUDA symbols to avoid segfaulting.

Kerilk:
To discuss...


TApplencourt:
Look good.

We initialize all your function pointers to `cuda_uninit`, this will avoid thapi crash if people code is wrong (calling a function without init first who is not part of implicit init).
And if we can a function that fail, we return `UNSUPORTED`, better than to crash indeed.

TApplencourt (dismissed):

TApplencourt (approved):

PR #147, opened 2023-09-08
Add subshell

TApplencourt:


TApplencourt:
The trap still seem to work, as the error code are propagated from the subshell to the parent

PR #148, opened 2023-09-11
Smaller energy thapi

TApplencourt:
This use `babeltrace_thapi` to replace `babelatrace_energy`. 

Handle it as a a normal event. We can add new option to the CLI to filter it or whatever if you want. 

Please tests is Solomon, I didn't...


Kerilk:
You would need to use it with `--backend=[]` right?
On a side note, why is it called `banckend` when it takes an array? Shouldn't it be called `--backends`?

Kerilk:
No, I meant thapi_energy was only displaying the energy none of the other events

TApplencourt:
Yeah, this is what I wrote on the description. I can add an option if you'd like to. But I guess you can grep the output for now

If we want to be general, we want to add an option to visualize only specific regex or whatever. Aka adding a `filter` before the pretty print sink.

Kerilk:
Ho yeah, hadn't seen your comment in the code. Let wait for @sbekele81 feedback then.

sbekele81:
I'm getting the following error when I ran the script: /home/sbekele/thapi-bin-new/bin/babeltrace_thapi:191:in `block (3 levels) in get_components': no implicit conversion of Array into String (TypeError)
	from /home/sbekele/thapi-bin-new/bin/babeltrace_thapi:177:in `each'
	from /home/sbekele/thapi-bin-new/bin/babeltrace_thapi:177:in `block (2 levels) in get_components'
	from /soft/packaging/spack/gnu-ldpath/build/linux-sles15-x86_64/gcc-11.2.0/ruby-3.1.0-55x7hesrduyvjspadyesab5n4llbevol/lib/ruby/gems/3.1.0/gems/babeltrace2-0.1.4/lib/babeltrace2/graph/graph.rb:114:in `block in _wrap_graph_simple_sink_component_consume_func'
	from /soft/packaging/spack/gnu-ldpath/build/linux-sles15-x86_64/gcc-11.2.0/ruby-3.1.0-55x7hesrduyvjspadyesab5n4llbevol/lib/ruby/gems/3.1.0/gems/babeltrace2-0.1.4/lib/babeltrace2/graph/graph.rb:452:in `bt_graph_run'
	from /soft/packaging/spack/gnu-ldpath/build/linux-sles15-x86_64/gcc-11.2.0/ruby-3.1.0-55x7hesrduyvjspadyesab5n4llbevol/lib/ruby/gems/3.1.0/gems/babeltrace2-0.1.4/lib/babeltrace2/graph/graph.rb:452:in `run'
	from /home/sbekele/thapi-bin-new/bin/babeltrace_thapi:368:in `<main>'
	from component-class-sink-simple.c:181.
	from graph.c:469.
	from /soft/packaging/spack/gnu-ldpath/build/linux-sles15-x86_64/gcc-11.2.0/ruby-3.1.0-55x7hesrduyvjspadyesab5n4llbevol/lib/ruby/gems/3.1.0/gems/babeltrace2-0.1.4/lib/babeltrace2/graph/graph.rb:455:in `run'
	from thapi-bin-new/bin/babeltrace_thapi:368:in `<main>'


babeltrace_energy calculates the delta energy and generates the power series.

TApplencourt:
May work now.

>As you mentioned, babeltrace_energy calculates the delta energy and generates the power series. if we are going to include it in the babeltrace_thapi, it would be nice that we have an option to filter.

Do you think  a `grep` is sufficient? If not the general solution is to create a new `regex` filter. 
We can also add `--energy-only` option in the tally but I don't like to add a new options as it's always hard to generate

sbekele81:
I think grep might do it. However,  the error still persists.

TApplencourt:
Fix now

sbekele81 (approved):

PR #149, opened 2023-09-11
Fix segfault when too many pluging are loaded

TApplencourt:


TApplencourt:
Tested passed, we can merge!

Kerilk (approved):

PR #155, opened 2023-10-10
APIs need to be in reverse order of their versions.

Kerilk:
`cuGetProcAddress` version where in incorrect order in the header leading to only `cuGetProcAddress_v2` being traced. This enforces the correct ordering.


TApplencourt (approved):

TApplencourt:
do you want something better in the sort in case one day they have v10? :p

PR #160, opened 2023-11-07
Btx ze

TApplencourt:


TApplencourt:
Before merging todo list:

- Fix slowdown in `metababel`
- Investigate difference in behavior between old version and this one for `~/iprof-20230717-205304/`

TApplencourt:
Ready to review @Kerilk  and @DonAurelio

utils/babeltrace_to_metababel_model.rb
```diff
@@ -1,17 +1,22 @@
...
-  'hipIpcEventHandle_t',
-  'hipPitchedPtr',
-  'hipExtent',
-  'dim3',
-]
+raise "Not input model provided" unless ARGV.length > 0
+input_model_path = ARGV[0]
+model = YAML.load_file(input_model_path)
+
+
```
DonAurelio, line 10:
```suggestion
```

ze/Makefile.am
```diff
@@ -250,68 +277,26 @@ data_DATA = \
...
-	babeltrace_zeinterval_callbacks.cpp \
-	babeltrace_zeprofiling_apis.txt \
-	babeltrace_zeinterval_callbacks.h \
-	babeltrace_zeinterval_dispatchers.c \
-	babeltrace_zeinterval_emitter.h \
-	babeltrace_zeinterval.h \
-	babeltrace_zeinterval.c \
+	$(BTX_ZE_GENERATED) \
 	xprof_utils.hpp \
 	xprof_utils.cpp
```
DonAurelio, line 286:
```suggestion
	xprof_utils.hpp
```

ze/btx_zeinterval_callbacks.cpp
```diff
@@ -0,0 +1,727 @@
+#include "xprof_utils.hpp"
+#include <btx_zeinterval_callbacks.hpp>
```
DonAurelio, line 2:
```suggestion
#include "btx_zeinterval_callbacks.hpp"
```

ze/btx_zeinterval_callbacks.cpp
```diff
@@ -0,0 +1,727 @@
...
+  else if (source == "lttng_ust_ze:zeMemAllocDevice_exit")
+    mi = &state->rangeset_memory_device;
+  else if (source == "lttng_ust_ze:zeMemAllocShared_exit")
+    mi = &state->rangeset_memory_shared;
+  else
+    std::cout << "WARNING Adding unknow memory " << source << "ptr " << ptr
+              << std::endl;
+
+  (*mi)[hp][ptr] = ptr + size;
+}
```
DonAurelio, line 37:
```suggestion
static void add_memory(data_t *state, hp_t hp, uintptr_t ptr, size_t size,
                       std::string source) {
  switch(source) {
    case "lttng_ust_ze:zeMemAllocHost_exit":
      state->rangeset_memory_host[hp][ptr] = ptr + size;
      break;
    case "lttng_ust_ze:zeMemAllocDevice_exit":
      state->rangeset_memory_device[hp][ptr] = ptr + size;
      break;
    case "lttng_ust_ze:zeMemAllocShared_exit"":
      state->rangeset_memory_shared[hp][ptr] = ptr + size;
      break;
    default:
      std::cout << "WARNING Adding unknown memory " << source << "ptr " << ptr << std::endl;
  }
}
```

DonAurelio: I think (not 100% sure) in C++ the switch should be able to manage string comparison. Also, I did not understand why you needed a temporary variable `m`, since you can update the values directly, perhaps I am wrong.

TApplencourt: Don't wnt to duplicate `  (*mi)[hp][ptr] = ptr + size;`

ze/btx_zeinterval_callbacks.cpp
```diff
@@ -0,0 +1,727 @@
...
+
+// Hash UUID
+static uintptr_t hash_device(const ze_device_properties_t &device_property) {
+  const auto device_uuid = device_property.uuid;
+
+  // Right now we store `ptr` and not `uuid`.
+  // we cast `uiid` to `ptr`. This should to the opposite
+  uintptr_t device_uuid_hash[2];
+  memcpy(&device_uuid_hash[0], &device_uuid, sizeof(device_uuid));
+  device_uuid_hash[0] += device_uuid_hash[1];
```
DonAurelio, line 137: You copy the value of `device_uuid` into `device_uuid_hash[0]`, but I do not see where `device_uuid_hash[1]` get a valid value.

ze/btx_zeinterval_callbacks.cpp
```diff
@@ -0,0 +1,727 @@
...
+  bool err = zeResult != ZE_RESULT_SUCCESS && zeResult != ZE_RESULT_NOT_READY;
+  btx_push_message_lttng_host(
+      btx_handle, hostname, vpid, vtid, start, BACKEND_ZE,
+      strip_event_class_name(event_class_name).c_str(), (ts - start), err);
+}
+
+/*
+ *
+ */
+
```
DonAurelio, line 182:
```suggestion
```

ze/btx_zeinterval_callbacks.cpp
```diff
@@ -0,0 +1,727 @@
...
+  return 0;
+}
+
+static bool memory_include(const memory_interval_t &m, const uintptr_t val) {
+  // if m empty, begin == end
+  const auto it = m.upper_bound(val);
+  if (it == m.cbegin())
+    return false;
+  return (val < (std::prev(it)->second));
+}
```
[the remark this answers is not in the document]
TApplencourt: We want to find something who is in the range, not the actual val. For example the interval can be [0...100] and we are locking to 10

ze/btx_zeinterval_callbacks.cpp
TApplencourt: `  memcpy(&device_uuid_hash[0], &device_uuid, sizeof(device_uuid));`  `device_uuid` have the size of `device_uuid_hash+device_uuid_hash` I think (devuce_uuid is a tuple if I remember correcly)

TApplencourt:
Everything look ok on my side.  @DonAurelio  can you approve  so we can merge? :) Thanks!

ze/btx_zeinterval_callbacks.hpp
```diff
@@ -0,0 +1,115 @@
...
+      device_timestamps_pair_ref;
+  std::unordered_map<
+      hp_event_t, std::tuple<thread_id_t, thapi_function_name, std::string,
+                             thapi_device_id, uint64_t, clock_lttng_device_t>>
+      event_payload;
+};
+typedef struct data_s data_t;
+
+// Push Pop. Verify that we never missuse them
+template <class S>
```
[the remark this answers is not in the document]
bd4: I guess this is for when is refactored to a common header or lib, since the data_t will be different for CUDA/HIP? Wondering if there is a nice way to limit/document the family of types that is designed to work with. I think is fine as a start for this PR.

bd4 (approved):
This is fairly complex code but it's doing a lot and I don't think there is much way around it. Things like the memory tracking could be improved with a high level tracker abstraction, with three instances for shared/device/host, but that can be part of refactoring when using for CUDA and HIP backend. My inclination is to merge this as is, to avoid future merge conflicts and to facilitate adding similar features to other backends and refactoring on the way as part of those PRs.

ze/btx_zeinterval_callbacks.hpp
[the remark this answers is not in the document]
TApplencourt: >

I guess this is for when is refactored to a common header or lib, since the data_t will be different for CUDA/HIP? Wondering if there is a nice way to limit/document the family of types that is designed to work with. I think is fine as a start for this PR.

 Exactly. Maybe we some fancy c++20 trait / concept, but for now I just templated. We can always pass `last-command` and get read of the `data_t` template

ze/btx_zeinterval_callbacks.hpp
```diff
@@ -0,0 +1,115 @@
...
+  std::unordered_map<hp_device_t, clock_lttng_device_t>
+      device_timestamps_pair_ref;
+  std::unordered_map<
+      hp_event_t, std::tuple<thread_id_t, thapi_function_name, std::string,
+                             thapi_device_id, uint64_t, clock_lttng_device_t>>
+      event_payload;
+};
+typedef struct data_s data_t;
+
+// Push Pop. Verify that we never missuse them
```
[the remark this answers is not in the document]
TApplencourt: Oh yeah, need to run `https://github.com/crate-ci/typos` on this thing!

ze/Makefile.am
```diff
@@ -36,6 +36,31 @@ MODIFIED_ZE_HDR = \
...
+	btx_filter_ze/metababel/btx_downstream.c \
+	btx_filter_ze/btx_main.c
+
+EXTRA_DIST = \
+	$(top_srcdir)/utils/babeltrace_to_metababel_model.rb \
+	$(top_srcdir)/xprof/btx_interval_model.yaml \
+	btx_zematching_model.yaml \
+	babeltrace_zeprofiling_apis.txt
+
+CLEANFILES = btx_ze_model.yaml
```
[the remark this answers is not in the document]
TApplencourt: The rules of thumb if that `dist-check didn't fail, it should be good.` In this particular case, they are added to line 283.

```
CLEANFILES += \
	$(BTX_ZE_GENERATED) \
	xprof_utils.hpp
```

So yes ;)

Kerilk (approved):

PR #164, opened 2024-01-05
[draft] port cuda filter to metababel

bd4:


bd4:
This tries to build metababel btx_*.c files with g++ compiler and fails to build.

cuda/btx_cuda_model.yaml
TApplencourt, line 1: Don't commit this guy please. It's generated

cuda/btx_cudamatching_model.yaml
TApplencourt, line 1: Should be remove in the futur :)

TApplencourt: (didn't read your comment oupsy)

cuda/btx_cuda_model.yaml
bd4: removed now

bd4:
Note that this has been rebased on `btx_ze` PR, which should be merged shortly.

TApplencourt:
@Kerilk  merged `btx_ze` 🎉 . You can rebase :)

bd4:
Rebased. Luckily I kept around the pre-rebased branch and just had to cherry pick my newest commits. One of the downsides of squashing is that it breaks rebase on a branch already rebased before the squash. Need someone to approve the workflow now.

TApplencourt:
Before I forgot: Please also modify the `configure` to add the new requirement on the bumped version of metababel

cuda/btx_cudamatching_model.yaml
```diff
@@ -0,0 +1,58 @@
...
+        :field_class:
+          :type: integer_signed
+          :cast_type: ^CUresult$
+
+  - :set_id: exits_cudaError_absent
+    :domain: exits - exits_cudaError_present 
+
+  # Memory traffic
+  - :set_id: traffic
+    :name: "cuMemcpy(2D|3D)?(Unaligned)?(DtoH|DtoD|HtoD)(Aync)?_v2(_ptds|_ptsz)?"
```
TApplencourt, line 39: Doesn'y `^cuMemcpy` work? or you are afraid of some of the memcopy should be ignored?

cuda/btx_cudamatching_model.yaml
```diff
@@ -0,0 +1,58 @@
...
+    :register: false
+  - :set_id: entries_traffic
+    :domain: entries & traffic
+    :payload_field_class:
+      :type: structure
+      :members:
+      - :name: ^ByteCount$
+        :field_class:
+          :cast_type: size_t
+          :type: integer_unsigned
```
TApplencourt, line 49: Does some `ByteCount` are NOT size_t? If not, you can remove the field_class

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+  send_host_message(btx_handle, usr_data, ts, event_class_name, hostname, vpid, vtid, false);
+}
+
+void exits_callback_cudaError_present(void *btx_handle, void *usr_data, int64_t ts,
+                                      const char *event_class_name, const char *hostname,
+                                      int64_t vpid, uint64_t vtid, CUresult cuResult) {
+
+  // Not an Error (cuResult == cudaErrorNotReady)
+  bool err = (cuResult != CUDA_SUCCESS) && (cuResult != CUDA_ERROR_NOT_READY);
```
TApplencourt, line 74: Please reuse `cuResultIsError` during futher refactoring

cuda/btx_cudamatching_model.yaml
bd4: I guess this is a question of failure modes and of documentation. I like being more explicit to document the type in the matching file. But then I guess if there was a ByteCount that was not size_t, the error would not be caught and the event would pass through the filter unchanged?

TApplencourt: No, it will be a compile-time error. As the `ByteCount` will have two types, hence two function signatures for the callbacks, which is obvious not possible.

bd4:
It will have compile time error if it does not match cast/type as you are suggesting. As I have it now, I think it would just match the one type and not filter the other, if another existed. right?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

bd4:
I can't reproduce the CI failure, not sure what is going on here. That file definitely was checked in and exists.

bd4:
The new test cases are very cuda specific, related to the context API. Probably should add `cuda` to the name and remove the ifs. The kernel_name and multithread cases could be made generic with some more testing.

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+#include <metababel/metababel.h>
+
+
+struct data_s {
+  // store entry state per thread in order to use in exit callback, conditioned
+  // on the cuResult being a non-error
+  std::unordered_map<hpt_t, int64_t> entry_ts;
+  std::unordered_map<hpt_t, std::string> entry_kernel_name;
+  std::unordered_map<hpt_t, size_t> entry_traffic_size;
```
TApplencourt, line 21: Can we use the generic `push / pop` void byte to save those one?

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+  // kernels (CUfunction objects) are "created" before use, at which time we save
```
TApplencourt, line 23: Maybe something like "Used to link CUfunction pointer to their ascii name". Maybe the rest is obvious? CuLaunch take a function pointer not an ascii name, so we need to store this info somewhere.

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  // the name so at execute time we know the name and can emit an event with the name
+  std::unordered_map<hp_kernel_t, std::string> hp_kernel_to_name;
+
+  // At device operation launch time, save the device the operation was launched on,
```
TApplencourt, line 27: It's not a "device operation", everything here is running on the host.
Some maybe "Used to save the last device in with the particular function have been launched".

And I guess "function" is bad, we should have use "command". More meaning full I think

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  // The flow of events for event profiling is:
+  //
+  // 1. device operation entry (cuMemcpy, cuLaunch, etc)
+  // 2. event_profiling
+  // 3. device operation exit
+  // 4. unknown number of intervening events
+  // ...
+  // 5. event profiling results
+  // 
+  // The information needed to emit a thapi device event is spread out throughout
```
TApplencourt, line 42: maybe use the full name of the yaml? Like `lttng::interval_device`? We will do a big sed when we will use correct name for the THAPI babeltrace2 event :)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  // these calls, so we need to maintain state between them.
+  // 
+  // Since 1 and 2 always happen one after another in a given thread, we just need
```
TApplencourt, line 45: If you use (2) after, please use (1) and (2) :)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+using data_t = struct data_s;
+
+void set_entry_ts(data_t *data, hpt_t hpt, int64_t ts) {
+  data->entry_ts[hpt] = ts;
+}
+
+int64_t get_entry_ts(data_t *data, hpt_t hpt) {
+  return data->entry_ts.at(hpt);
+}
```
TApplencourt, line 64: Do we need a helper function for such a simple task?

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+};
+
+using data_t = struct data_s;
+
+void set_entry_ts(data_t *data, hpt_t hpt, int64_t ts) {
+  data->entry_ts[hpt] = ts;
+}
+
+int64_t get_entry_ts(data_t *data, hpt_t hpt) {
+  return data->entry_ts.at(hpt);
```
TApplencourt, line 63: Not sure about an `at` in a performance critical place like this. Bound checking have a place. So maybe we should `ifdef` this guy so that we we are in debug more we use `at` and in release mode we use `[]`? 

I guess I'm just asking for benchmarking, but yeah I'm a little afraid of performance impact

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  std::unordered_map<hpt_t, fn_ts_t> hpt_profiled_function_name_and_ts;
+  std::unordered_map<hp_event_t, fn_ts_t> hp_event_to_function_name_and_ts;
+
+  // Encapsulate complex context management state
+  CUDAContextManager context_manager;
+};
+
+using data_t = struct data_s;
+
+void set_entry_ts(data_t *data, hpt_t hpt, int64_t ts) {
```
TApplencourt, line 58: everything should be static so we don't polute the symbols. Or anonymous namespace if you prefer

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+                              event_class_name_striped.c_str(), (ts - _start), err);
+}
+
+static void send_traffic_message(void *btx_handle, void *usr_data, int64_t ts,
+                                 const char *event_class_name, const char *hostname,
+                                 int64_t vpid, uint64_t vtid, bool err) {
+  std::string event_class_name_stripped = strip_event_class_name(event_class_name);
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t key = {hostname, vpid, vtid};
+  const int64_t _start = get_entry_ts(state, key);
```
TApplencourt, line 94: In `send _host` we don't use `state`, so please pick one. ( i prefer the smalest one :) )

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+  if (!err) {
+    auto size = state->entry_traffic_size.at(key);
```
TApplencourt, line 97: Same can remove at. We should always have it. A user cannot write a code where we don't have it.

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+void btx_initialize_component(void **usr_data) { *usr_data = new data_t; }
+
+void btx_finalize_component(void *usr_data) {
+  delete static_cast<data_t *>(usr_data);
+}
+
+void entries_callback(void *btx_handle, void *usr_data, int64_t ts,
+                      const char *event_class_name, const char *hostname, int64_t vpid,
+                             uint64_t vtid) {
+  set_entry_ts(((data_t *)usr_data), {hostname, vpid, vtid}, ts);
```
TApplencourt, line 113: Yeah I think we can inline `set_entry_ts` :)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+                                     int64_t vpid, uint64_t vtid) {
+  hpt_t hpt = {hostname, vpid, vtid};
+  try {
+    send_host_message(btx_handle, usr_data, ts, event_class_name,
+                      hostname, vpid, vtid, false);
+  } catch(const std::out_of_range& oor) {
+    THAPI_FATAL_HPT(hpt,
+      "entry ts not found for thread in exit cudaError_absent callback");
+  }
+}
```
TApplencourt, line 127: Cannot append, so can remove some code :)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  set_entry_ts(state, key, ts);
+  state->entry_traffic_size[key] = size;
+}
+
+void entries_traffic_v1_callback(void *btx_handle, void *usr_data, int64_t ts,
+                                 const char *event_class_name, const char *hostname,
+                                 int64_t vpid, uint64_t vtid, uint32_t size) {
+  entries_traffic_v2_callback(btx_handle, usr_data, ts, event_class_name, hostname,
+                              vpid, vtid, static_cast<size_t>(size));
+}
```
TApplencourt, line 159: They have the same function signature. So in theory can be merge in one callback in  the `matching` yaml? and some fancy `cast_type` (it's ok to cast `uint32_t` to `size_t`)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  try {
+    send_host_message(btx_handle, usr_data, ts, event_class_name,
+                      hostname, vpid, vtid, err);
+  } catch(const std::out_of_range& oor) {
+    THAPI_FATAL_HPT(hpt,
+                    "entry ts not found for thread in exit cudaError_present callback");
+  }
+}
+
+void entries_traffic_v2_callback(void *btx_handle, void *usr_data, int64_t ts,
```
TApplencourt, line 144: Please add a  comment to say that the only difference at the type of the size :)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  // Not an Error (cuResult == cudaErrorNotReady)
+  bool err = (cuResult != CUDA_SUCCESS) && (cuResult != CUDA_ERROR_NOT_READY);
+  try {
+    send_traffic_message(btx_handle, usr_data, ts, event_class_name,
+                         hostname, vpid, vtid, err);
+  } catch(const std::out_of_range& oor) {
+    THAPI_FATAL_HPT(hpt,
+                    "entry traffic size not found for hpt in traffic message");
+  }
+}
```
TApplencourt, line 175: Same can remove try /cast as  size will always exist (set by the entry)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  try {
+    hpt_function_name_t hpt_function_name = {hostname, vpid, vtid, fn_name};
+    auto dev = state->hpt_function_name_to_dev.at(hpt_function_name);
+    bool err = cuResultIsError(startStatus) || cuResultIsError(stopStatus)
+             || cuResultIsError(status);
+
+    uint64_t delta = static_cast<uint64_t>(ms * 1e6);
+    if (err) {
+      delta = 0;
+    }
```
TApplencourt, line 247: Can use a ternary
```suggestion
    uint64_t delta =  err ? 0 : static_cast<uint64_t>(ms * 1e6);
```

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+                                      uint64_t vtid, CUdevice dev) {
+  auto state = static_cast<data_t *>(usr_data);
+  state->context_manager.primary_ctx_reset_entry({hostname, vpid, vtid}, dev);
+}
+
+void primary_ctx_reset_exit_callback(void *btx_handle, void *usr_data,
+                                     int64_t ts, const char *event_class_name,
+                                     const char *hostname, int64_t vpid,
+                                     uint64_t vtid, CUresult cuResult) {
+  auto state = static_cast<data_t *>(usr_data);
```
TApplencourt, line 306: Should we handle the `cu_result` outside of the Context_manager? 
It will mimic what we do for other function. Not any strong preference

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  state->context_manager.ctx_create_exit({hostname, vpid, vtid}, cuResult, ctx);
+}
+
+void ctx_destroy_entry_callback(void *btx_handle, void *usr_data,
+                                int64_t ts, const char *event_class_name,
+                                const char *hostname, int64_t vpid,
+                                uint64_t vtid, CUcontext ctx) {
+  auto state = static_cast<data_t *>(usr_data);
+  state->context_manager.ctx_destroy_entry({hostname, vpid, vtid}, ctx);
+}
```
TApplencourt, line 333: Should we replace all those stuff by some MACRO? If the name match should be doable.

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+// Kernel name and device tracking
+// ===============================
+void module_get_function_entry_callback(void *btx_handle, void *usr_data,
+                                        int64_t ts, const char *event_class_name,
+                                        const char *hostname, int64_t vpid,
+                                        uint64_t vtid, char* name) {
+  std::string name_str(name);
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t hpt = {hostname, vpid, vtid};
+  state->entry_kernel_name[hpt] = name_str;
```
TApplencourt, line 415:
```suggestion
  state->entry_kernel_name[{hostname, vpid, vtid}] = name_str;
```

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+}
+
+void module_get_function_exit_callback(void *btx_handle, void *usr_data,
+                                       int64_t ts, const char *event_class_name,
+                                       const char *hostname, int64_t vpid,
+                                       uint64_t vtid, CUresult cuResult,
+                                       CUfunction cuFunction) {
+  if (cuResultIsError(cuResult)) {
```
TApplencourt, line 423: Oh so you still have `cuResultIsError`, please use it everywhere :)

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+                            int64_t ts, const char *event_class_name,
+                            const char *hostname, int64_t vpid,
+                            uint64_t vtid, CUstream cuStream, CUfunction f) {
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t hpt = {hostname, vpid, vtid};
+  CUdevice dev;
+  bool found = state->context_manager.get_stream_device(hpt, cuStream, &dev);
+  if (!found) {
+    THAPI_FATAL_HPT(hpt, "no device for thread in device op");
+  }
```
TApplencourt, line 448:
```suggestion
  bool found = state->context_manager.get_stream_device(hpt, cuStream, &dev);
  if (!found) {
    THAPI_FATAL_HPT(hpt, "no device for thread in device op");
  }
```
```suggestion
 if (state->context_manager.get_stream_device(hpt, cuStream, &dev)
    THAPI_FATAL_HPT(hpt, "no device for thread in device op");
```

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+  std::string name = strip_event_class_name(event_class_name);
+  if (f == nullptr) {
```
TApplencourt, line 451: I think we should use `std::optional` then

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+  } else {
+    // for device launch, get the name saved when the kernel was loaded in the
+    // module get function callbacks
+    try
+    {
+      std::string name = state->hp_kernel_to_name.at({hostname, vpid, f});
+      state->hpt_function_name_to_dev[{hostname, vpid, vtid, name}] = dev;
+      state->hpt_profiled_function_name_and_ts[hpt] = fn_ts_t(name, ts);
+    } catch(const std::out_of_range& oor) {
+      THAPI_FATAL_HPT(hpt,
```
TApplencourt, line 465: The only difference between the 2 code path seem to be how we can the name. 

Please make it more obvious. Aka a `if / else` to get the name, then the `hpt_function_name_to_dev` and `hpt_profiled_function_name_and_ts` outside the block

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+        "device not found for hpt function name in kernel launch callback");
+    }
+  }
+}
+
+void named_operation_stream_present_entry_callback(void *btx_handle, void *usr_data,
+                                                   int64_t ts, const char *event_class_name,
+                                                   const char *hostname, int64_t vpid,
+                                                   uint64_t vtid, CUstream cuStream) {
+  operation_entry_helper(btx_handle, usr_data, ts, event_class_name,
```
TApplencourt, line 475: Not a good name, not really explicit

TApplencourt (changes requested):

cuda/btx_cudainterval_callbacks.cpp
bd4: Ya I think we can remove the helper, it is not really making code significantly shorter or easier to understand. The getter could be useful in having .at vs [] depending on if Debug or Release build, but that is maybe better done with a wrapper or subclass of std::map.

cuda/btx_cudainterval_callbacks.cpp
bd4: So see if metababel lets us get away with having all of them (with uint32_t or size_t params) go to a single callback with size_t? I will try.

cuda/btx_cudainterval_callbacks.cpp
bd4: Discussed this a lot on slack, decided to drop verbose try/catch error handling in favor of THAPI_AT macro along with valgrind to get stack traces.

cuda/btx_cudainterval_callbacks.cpp
bd4: I can see the logic of doing it outside context manager, but don't have a strong preference either, so inclined to leave it as is.

cuda/btx_cudainterval_callbacks.cpp
bd4: I changed to use `task` and added some comments, let me know if new code is clear enough.

cuda/btx_cudainterval_callbacks.cpp
bd4: Makes line too long to put inside the if, and I like the clarity of found name.

bd4: Also this is the one place I left THAPI_FATAL. Not sure if worth keeping the macros or keeping this check. May need a perf check to see if having context manager check this slows it down at all. Context manager is at least called a lot less then e.g. entry state, i.e. on per task basis, not on per callback bases. And code that does lots of small kernel launches is generally discouraged.

cuda/btx_cudainterval_callbacks.cpp
bd4: I don't love macros for this. It's ugly but clear now? Are you thinking something like this:

`#define STATE_CAST(data)  static_cast<data_t *>(data)`

and

`auto state = STATE_CAST(usr_data)`

Not totally sure macros even work for static cast, I guess it should.

cuda/btx_cudainterval_callbacks.cpp
bd4: I kept the same callbacks and added comments to make it more clear, let me know what you think.

cuda/btx_cudainterval_callbacks.cpp
bd4: Also use "kernel task" and "non kernel task" terminology to distinguish the two, I think is a lot clearer. The kernel ones have the CUfunction arg, everything else does not and uses the task launch function name for the emitted event name.

cuda/entry_state.hpp
```diff
@@ -0,0 +1,77 @@
...
+            typename = std::enable_if_t<std::is_same_v<K, std::string>>>
+  std::enable_if_t<std::is_same_v<K, std::string>, K>
+  pop_entry(hpt_t hpt) {
+    auto &v = pop_entry_impl(hpt);
+    const std::string res{(char *)v.value().data()};
+    v.reset();
+    return res;
+  }
+
+  void set_ts(hpt_t hpt, int64_t ts) {
```
TApplencourt, line 49: I guess we should choose between `pop / push` and `set / get`. They are doing the same thing.

I vote maybe more for `set / get` , as we cannot push multiple entry for the `hpt`, it will be overwritten.

cuda/btx_cudainterval_callbacks.cpp
TApplencourt: If the logic of our code is correct, this should never happen.

If this happend, I guess we can always put back some tests, (or wait for the `.at()` improvement / debug)

cuda/btx_cudamatching_model.yaml
```diff
@@ -0,0 +1,293 @@
...
+    :domain: exits & module_get_function
+    :payload_field_class:
+      :type: structure
+      :members:
+      - :name: ^cuResult$
+      - :name: ^hfunc_val$
+
+  # tasks that generate thapi device events
+  - :set_id: task
+    :name: "(cuMemcpy|cuMemPrefetch|cuMemset|cuMemFree|cuLaunch)"
```
TApplencourt, line 240: nitpicking but I think  the `()` are spurious

TApplencourt: Maybe just `cuMem` ? You do that for `cuLaunch`

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
+#include "context_manager.hpp"
+
+
+bool CUDAContextManager::get_device(hpt_t hpt, CUdevice *pdev) {
+  CUcontext ctx;
+  bool found = get_top_context(hpt, &ctx);
+  if (!found) {
+    return false;
+  }
+  // std::cerr << "get device for thread " << std::get<2>(hpt) << std::endl;
```
TApplencourt, line 10: Please remove commant, we will put them back if needed :)

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+  *pctx = hpt_stack.top();
+  return true;
+}
+
+bool CUDAContextManager::get_stream_device(hpt_t hpt, CUstream stream, CUdevice *pdev) {
+  if (stream == 0) {
+    // using the default stream in the top context on the stack
+    return get_device(hpt, pdev);
+  }
+  hp_stream_t hp_stream_key = {std::get<0>(hpt), std::get<1>(hpt), stream};
```
TApplencourt, line 37: I think it's "good practice" to remove the `=`

`hp_stream_t hp_stream_key{std::get<0>(hpt), std::get<1>(hpt), stream};`

should work

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+  CUcontext ctx;
+  ctx = THAPI_AT(hp_stream_to_ctx_, hp_stream_key);
+  auto ctx_find = hp_ctx_to_device_.find({std::get<0>(hpt), std::get<1>(hpt), ctx});
```
TApplencourt, line 40: you are use `hp` twice, maybe worth putting it in a variable?

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+void CUDAContextManager::primary_ctx_retain_exit(hpt_t hpt, CUresult cuResult,
+                                                 CUcontext ctx) {
+  if (cuResultIsError(cuResult)) {
+    return;
+  }
+  // std::cerr << "primaryctx retain exit thread " << std::get<2>(hpt) << std::endl;
+  try {
+    auto entry_dev = THAPI_AT(hpt_entry_dev_, hpt);
```
TApplencourt, line 61: need `&` or not?

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+    auto ctx = THAPI_AT(hp_device_primary_ctx_, hp_dev_key);
+    hp_device_primary_ctx_.erase(hp_dev_key);
+    hp_ctx_to_device_.erase({std::get<0>(hpt), std::get<1>(hpt), ctx});
+  } catch(const std::out_of_range& oor) {
+    THAPI_FATAL_HPT(hpt,
+                    "no context found in primary ctx release exit callback");
+  }
+}
+
+
```
TApplencourt, line 99: Why sometime 2 space, sometime 1?

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+// cuPrimaryCtxReset_v2_(entry|exit)
+void CUDAContextManager::primary_ctx_reset_entry(hpt_t hpt, CUdevice dev) {
+  primary_ctx_release_entry(hpt, dev);
+}
```
TApplencourt, line 103: why `reset` and then `release`?

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+void CUDAContextManager::primary_ctx_reset_exit(hpt_t hpt, CUresult cuResult) {
+  primary_ctx_release_exit(hpt, cuResult);
+}
+
+
+// cuCtxCreate_v(2|3)_(entry|exit)
+void CUDAContextManager::ctx_create_entry(hpt_t hpt, CUdevice dev) {
+  hpt_entry_dev_[hpt] = dev;
```
TApplencourt, line 112: Should we use an internal the newly introduced `EntryState`  to do so?

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+  }
+  hp_stream_t hp_stream_key = {std::get<0>(hpt), std::get<1>(hpt), cuStream}; 
+  
+  hp_stream_to_ctx_[hp_stream_key] = ctx;
+}
+
+void CUDAContextManager::ensure_stack(hpt_t hpt) {
+  if (hpt_ctx_stack_.count(hpt) == 0) {
+    hpt_ctx_stack_[hpt] = {};
+  }
```
TApplencourt, line 257: What is this function?
It sounds like a code smell to me. 

`{}` is the default value, look like it emulate the behavior of `[]`?

cuda/context_manager.cpp
```diff
@@ -0,0 +1,258 @@
...
+                                            CUstream cuStream) {
+  if (cuResultIsError(cuResult)) {
+    return;
+  }
+  CUcontext ctx;
+  bool found = get_top_context(hpt, &ctx);
+  if (!found) {
+    // TODO: how to we raise error here? Should not reach
+  }
+  hp_stream_t hp_stream_key = {std::get<0>(hpt), std::get<1>(hpt), cuStream};
```
TApplencourt, line 249: I guesss no need for tmp variable

ze/Makefile.am
```diff
@@ -316,7 +316,8 @@ tests/ze_dummy_trace.yaml: $(top_srcdir)/utils/gen_dummy_trace.rb ze_babeltrace_
 #https://www.gnu.org/software/make/manual/html_node/Static-Usage.html
 
 TRACE_COMMON = \
-	tests/interval_profiling_normal.yaml
+	tests/interval_profiling_normal.yaml \
+	tests/interval_profiling_multithread.yaml
```
TApplencourt, line 320: Thanks!

cuda/entry_state.hpp
bd4: ya I like set / get. I wonder about the cost of doing .reset also, I guess it is just deleting a key not zeroing values so could be quite fast.

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+}
+
+void entries_traffic_v2_callback(void *btx_handle, void *usr_data, int64_t ts,
+                                 const char *event_class_name, const char *hostname,
+                                 int64_t vpid, uint64_t vtid, size_t size) {
+  // save traffic size and entry ts for use in exit callback
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t key = {hostname, vpid, vtid};
+  set_entry_ts(state, key, ts);
+  state->entry_traffic_size[key] = size;
```
TApplencourt: should we use the `EntryState`?

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,589 @@
...
+                              vpid, vtid, static_cast<size_t>(size));
+}
+
+void exits_traffic_callback(void *btx_handle, void *usr_data, int64_t ts,
+                            const char *event_class_name, const char *hostname,
+                            int64_t vpid, uint64_t vtid, CUresult cuResult) {
+  hpt_t hpt = {hostname, vpid, vtid};
+
+  // Not an Error (cuResult == cudaErrorNotReady)
+  bool err = (cuResult != CUDA_SUCCESS) && (cuResult != CUDA_ERROR_NOT_READY);
```
TApplencourt: Should we the `cuda_error_function`?

cuda/btx_cudainterval_callbacks.cpp
TApplencourt: Look good!

cuda/btx_cudainterval_callbacks.cpp
TApplencourt: No sorry was thinkinking for about 

```
DEFINE CONTEXT_MANAGER_CALLBACK_FOO


void $FOO_callback(...) {
   state->context_manage.$FOO(...)
}
```

cuda/btx_cudainterval_callbacks.cpp
TApplencourt: Nice job!

cuda/context_manager.cpp
bd4: Ya it was needed when using .at everywhere, which is still true for THAPI_AT which can be .at. But I think in this case I can use [] in the two places (ctx set and ctx create) where the stack can be created, and use AT elsewhere, and still have reasonable debugability.

cuda/context_manager.cpp
bd4: Good idea, I will see if I can easily pass it to the constructor or maybe set it after the fact in the filter init.

cuda/context_manager.cpp
bd4: Good question, the have different semantics in the API but same affect as far as we care about, but that is very much not obvious and I can't even remember why... so def needs comment or a better name for common routine or both.

cuda/context_manager.cpp
bd4: I think CUdevice, CUcontext, CUfunction, are all pointer types and it doesn't matter, but not sure how I feel about that, is not always obvious. Also we don't mutate them, so even if it had to copy a non small object, it would not break, just be extra copy. Inclined to never use &, definitely need to be consistent here. The one case we need it is for when the value type is a stack and we ARE mutating it... so I guess be consistent of only use it when we really need to mutate.

cuda/context_manager.cpp
bd4: is hp stream and context, not same key.

cuda/context_manager.cpp
bd4: Changed to use auto& for stack and no & everywhere else.

cuda/context_manager.cpp
bd4: added some comments.

cuda/context_manager.cpp
TApplencourt: Perfect!

cuda/context_manager.hpp
```diff
@@ -0,0 +1,135 @@
...
+  std::unordered_map<hp_context_t, CUdevice> hp_ctx_to_device_;
+
+  // Track streams so when they are passed to API calls we know what context
+  // (and thereby what device) the API call is using
+  std::unordered_map<hp_stream_t, CUcontext> hp_stream_to_ctx_;
+
+  // Temp saves to carry args from entry to exit (since we need to wait for exit
+  // callback and check cuResult before doing anything). Need to keep track per thread.
+  std::unordered_map<hpt_t, CUcontext> hpt_entry_ctx_;
+  std::unordered_map<hpt_t, CUdevice> hpt_entry_dev_;
```
TApplencourt: Should we use an `entry_state` for those one?

cuda/entry_state.hpp
TApplencourt: So let rename `push/ pop`  `set / get` then :)

cuda/context_manager.hpp
bd4: Not sure what is going on with github, but latest version uses entry_state and does not have these.

cuda/context_manager.cpp
```diff
@@ -0,0 +1,168 @@
...
+    return;
+  }
+  auto entry_ctx = entry_state_.get_data<CUcontext>(hpt);
+  // Note: don't use THAPI_AT, we need operator[] semantics of
+  // creating if the stack does not exists yet.
+  auto &stack = hpt_ctx_stack_[hpt];
+  if (!stack.empty()) {
+    stack.pop();
+  }
+  stack.push(entry_ctx);
```
TApplencourt, line 132: only if `entry_ctx` is non null

cuda/btx_cudainterval_callbacks.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include "xprof_utils.hpp"
+
+using hp_event_t = std::tuple<hostname_t, process_id_t, CUevent, CUevent>;
+using hp_kernel_t = std::tuple<hostname_t, process_id_t, CUfunction>;
+using hp_stream_t = std::tuple<hostname_t, process_id_t, CUstream>;
+using hpt_context_t =
+    std::tuple<hostname_t, process_id_t, thread_id_t, CUcontext>;
+using hp_context_t = std::tuple<hostname_t, process_id_t, CUcontext>;
+using context_stack_t = std::stack<CUcontext>;
```
TApplencourt, line 15: Can this should be in context manager? as it's used only there

Here I think we should put stuff used in the `cudainterval_callback.cpp`

cuda/btx_cudamatching_model.yaml
```diff
@@ -0,0 +1,293 @@
...
+      :members:
+      - :name: ^cuResult$
+
+  # Stream create
+  - :set_id: stream_create
+    :name: "cuStreamCreate"
+    :register: false
+  - :set_id: stream_create_entry
+    :domain: entries & stream_create
+    :register: false
```
TApplencourt, line 211: Can you just delete it , if you don't need to register?

cuda/btx_cudamatching_model.yaml
```diff
@@ -0,0 +1,293 @@
...
+      :members:
+      - :name: ^hStream$
+
+  - :set_id: task_stream_absent_entry
+    :domain: task_entry - task_stream_present_entry
+    :register: false
+
+  # specialize for whether a funtion arg is present or not
+  - :set_id: kernel_task_entry
+    :name: "cuLaunch[^H]"
```
TApplencourt, line 261: Maybe a comment tp explain why not the `H`? For Host i  guess :)

cuda/btx_cudamatching_model.yaml
```diff
@@ -0,0 +1,293 @@
...
+    :domain: task_entry
+    :register: false
+    :payload_field_class:
+      :type: structure
+      :members:
+      - :name: ^f$
```
TApplencourt, line 267: Not sure if you need that, because after you still do the `-name: ^f$`

utils/tests_template/interval_profiling_multithread.yaml.erb
```diff
@@ -0,0 +1,188 @@
...
+  - :name: lttng_ust_cuda:cuCtxSetCurrent_exit
+    :common_context:
+      vtid: 1
+  - :name: lttng_ust_cuda:cuCtxCreate_entry
+    :common_context:
+      vtid: 2
+  - :name: lttng_ust_cuda:cuCtxCreate_exit
+    :common_context:
+      vtid: 2
+  <% end %>
```
TApplencourt, line 58: Do you need to set some argument?

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,443 @@
...
+  btx_register_callbacks_lttng_ust_cuda_profiling_event_profiling(
+      btx_handle, &profiling_callback);
+  btx_register_callbacks_lttng_ust_cuda_profiling_event_profiling_results(
+      btx_handle, &profiling_callback_results);
+
+  // Context handling
+  btx_register_callbacks_primary_ctx_retain_entry(
+      btx_handle, &primary_ctx_retain_entry_callback);
+  btx_register_callbacks_primary_ctx_retain_exit(
+      btx_handle, &primary_ctx_retain_exit_callback);
```
TApplencourt, line 391: I guess we can create a MACRO too to avoid code bloat, and enforce consistent naming? Not important

cuda/btx_cudamatching_model.yaml
bd4: confirmed not needed

cuda/btx_cudamatching_model.yaml
bd4: added a comment, agreed it is not obvious

cuda/btx_cudamatching_model.yaml
bd4: yup kept it around "just in case" but clear now is not needed I think.

cuda/btx_cudainterval_callbacks.hpp
bd4: good call.

cuda/btx_cudainterval_callbacks.cpp
bd4: good idea! was quick to do.

cuda/btx_cudainterval_callbacks.cpp
```diff
@@ -0,0 +1,443 @@
...
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t hpt{hostname, vpid, vtid};
+  auto dev = state->context_manager.get_stream_device(hpt, cuStream);
+
+  std::string name = strip_event_class_name(event_class_name);
+  if (f_optional.has_value()) {
+    auto f = f_optional.value();
+    // for device launch, get the name saved when the kernel was loaded in the
+    // module get function callbacks
+    hp_kernel_t hp_kernel_key{hostname, vpid, f};
```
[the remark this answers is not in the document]
bd4: did not work but used only once, so `f_option.value()` is good.

bd4:
For my 11MB fft bench test case, master branch is ~2x faster than this feature branch. Not sure how much of this is from misc bug fixes vs actual regression.

bd4:
With my manual test cases, this is producing same results as master branch except for the expected differences:

1. Reverse order of host event and traffic event
2. Fix backend=3 for device events
3. Add ts to traffic event

It also runs successfully on MPI/CUDA apps, specifically a test app using gtensor and MPI.

I think the main potential blocker is that at least on one test case, it is 2x slower than master branch. We can merge and then try to address the performance, or hold off and try to improve it first.

TApplencourt (approved):

TApplencourt (approved):

PR #182, opened 2024-02-21
use ISO8601

TApplencourt:
`https://en.wikipedia.org/wiki/ISO_8601` the `:` make it a little painful but still manageable (no need to escape) 

At least now we will use a standard 🤷🏽


bd4 (dismissed):
Looks good!

bd4 (approved):
LGTM!

PR #189, opened 2024-03-28
Pr/opencl btx

bd4:


opencl/btx_clinterval_callbacks.cpp
```diff
@@ -213,11 +214,55 @@ static void launch_function_entry_callback(void *btx_handle, void *usr_data, int
...
 }
 
+static void profiling_callback(void *btx_handle, void *usr_data, int64_t ts,
+                               const char *hostname, int64_t vpid,
+                               uint64_t vtid, int status, cl_event event) {
+  auto state = static_cast<data_t *>(usr_data);
+  auto fn_name_ts = thapi_at(state->profiled_function_name_and_ts,
+                             hpt_t{hostname, vpid, vtid});
+  auto function_name = std::get<0>(fn_name_ts);
+  auto launch_ts = std::get<1>(fn_name_ts);
```
TApplencourt, line 229: can do `auto [_timestamp, launch_ts] = fn_name_ts;` I think :)

opencl/btx_openclmatching_model.yaml
```diff
@@ -0,0 +1,117 @@
+# Note: all API calls seen by THAPI are driver API calls, even when the
+# application uses only runtime API (since runtime API is just high level
+# wrapper around driver API).
+#
+# See also:
+#  https://docs.nvidia.com/cuda/cuda-driver-api/driver-vs-runtime-api.html
```
TApplencourt, line 6: Can be removed :)

opencl/gen_babeltrace_cl_model.rb
```diff
@@ -92,6 +92,47 @@ def parse_field(field)
...
+  {
+    name: 'vtid',
+    class: 'unsigned',
+    cast_type: 'uint64_t',
+    class_properties: {
+      field_value_range: 64,
+    }
+  }
+]
+
```
TApplencourt, line 131: It is not clear to me why they are now required where they were not before.

utils/bt_plugins/comparator.rb
```diff
@@ -29,6 +30,9 @@ def initialize_method(self_component, _configuration, _params, _data)
   end
 
   def verify_messages
+    if @stack_messages.values.map { |data| data.length }.to_set.length > 1
```
TApplencourt, line 33: Good catch! 

If they are not the same size, `transpose`should fail with an error. Then it possible to `rescue` to print a better error message if it was not clear enough.

PS: You can do `@stack_messages.values.uniq{ |data| data.length }.length  > 1 ` (so no need to rely on set :)

utils/tests_template/interval_profiling_kernel_name.yaml.erb
```diff
@@ -111,11 +111,7 @@ ref:
   - :name: lttng:device
     :common_context:
       hostname: aurora12.gov
-<% if ["ze", "cuda"].include?(be) %>
```
TApplencourt, line 114: Fill good! :D

opencl/gen_babeltrace_cl_model.rb
bd4: I thought it was related to btx conversion, all the other backends have them. But does opencl not have them?

opencl/btx_clinterval_callbacks.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include "xprof_utils.hpp"
+
+using hp_command_queue_t = std::tuple<hostname_t, process_id_t, cl_command_queue>;
+using hp_kernel_t = std::tuple<hostname_t, process_id_t, cl_kernel>;
+using hp_event_t = std::tuple<hostname_t, process_id_t, cl_event>;
+
+inline bool clRvalIsError(cl_int &cl_rval) {
+  return (cl_rval != CL_SUCCESS) && (cl_rval != CL_DEVICE_NOT_AVAILABLE);
+}
```
TApplencourt, line 15: why in the `hpp`? doesn't seem to be templated

utils/bt_plugins/comparator.rb
```diff
@@ -1,6 +1,7 @@
 require 'babeltrace2'
 require 'yaml'
 require 'optparse'
+require 'set'
```
TApplencourt, line 4: can be removed

opencl/Makefile.am
```diff
@@ -211,24 +232,6 @@ libBabeltraceCL_la_CPPFLAGS = -I$(top_srcdir)/utils -I$(top_srcdir)/utils/includ
 libBabeltraceCL_la_CFLAGS = -Wall -Wextra -Wno-unused-parameter $(WERROR) $(BABELTRACE2_CFLAGS)
 libBabeltraceCL_la_LDFLAGS = $(BABELTRACE2_LIBS) -avoid-version -module
 
-babeltrace_clinterval_callbacks.cpp: $(srcdir)/gen_clinterval_callbacks.rb $(srcdir)/clinterval_callbacks.cpp.erb opencl_babeltrace_model.yaml
```
TApplencourt, line 214: should remove `gen_clinterval_callbacks.rb` from the git

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+                                         const char *event_class_name,
+                                         const char *hostname, int64_t vpid,
+                                         uint64_t vtid, cl_int errcode_ret_val,
+                                         cl_uint num_devices_val,
+                                         cl_device_id *devices_vals) {
+  auto state = static_cast<data_t *>(usr_data);
+  if ((devices_vals != nullptr) && (errcode_ret_val == CL_SUCCESS)) {
+    for (unsigned int i=0; i < num_devices_val; i++) {
+      const thapi_device_id d = reinterpret_cast<thapi_device_id>(devices_vals[i]);
+      state->device_to_root_device[hp_device_t(hostname, vpid, d)] = d;
```
TApplencourt, line 125: can remove `hp_device_t` cast an rely of auto deduction

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+                                             cl_device_id *out_devices_vals) {
+  auto state = static_cast<data_t *>(usr_data);
+  if ((out_devices_vals != nullptr) && (errcode_ret_val == CL_SUCCESS)) {
+    hpt_t key{hostname, vpid, vtid};
+    auto in_device = state->entry_state.get_data<thapi_device_id>(key);
+    const thapi_device_id root_device = state->device_to_root_device[
+                                                  hp_device_t(hostname, vpid, in_device)];
+    for (unsigned int i=0; i < num_devices_ret_val; i++ ) {
+      const thapi_device_id d = reinterpret_cast<thapi_device_id>(out_devices_vals[i]);
+      state->device_to_root_device[hp_device_t(hostname, vpid, d)] = root_device;
```
TApplencourt, line 154: `{(hostname, vpid, d}`

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+                                               int64_t ts, const char *event_class_name,
+                                               const char *hostname, int64_t vpid,
+                                               uint64_t vtid,
+                                               cl_command_queue command_queue,
+                                               cl_int errcode_ret_val) {
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t key{hostname, vpid, vtid};
+  auto entry_device = state->entry_state.get_data<thapi_device_id>(key);
+  const auto root_device = state->device_to_root_device[
+                                              hp_device_t(hostname, vpid, entry_device)];
```
TApplencourt, line 178:
```suggestion
                                              {hostname, vpid, entry_device}];
```

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+  hp_command_queue_t cq_key{hostname, vpid, command_queue};
+  if (root_device != 0) {
```
TApplencourt, line 180: I think it cannot be `0` As we always add it to the map

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+    state->command_queue_to_device[cq_key] = dsd_t(entry_device, 0) ;
+  }
+}
+
+static void kernel_info_callback(void *btx_handle, void *usr_data, int64_t ts,
+                                 const char *event_class_name, const char *hostname,
+                                 int64_t vpid, uint64_t vtid, cl_kernel kernel,
+                                 char *function_name) {
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t key{hostname, vpid, vtid};
```
TApplencourt, line 192: Key is not used. (How that even compile with `-Wall`? :') )

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+  // TODO: do we need to save a copy?
+  state->kernel_to_name[hp_kernel_t(hostname, vpid, kernel)] = function_name;
```
TApplencourt, line 194:
```suggestion
  state->kernel_to_name[{hostname, vpid, kernel}] = function_name;
```

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+                                         const char *hostname, int64_t vpid,
+                                         uint64_t vtid, cl_command_queue command_queue,
+                                         cl_kernel kernel) {
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t key{hostname, vpid, vtid};
+  auto name = state->kernel_to_name[hp_kernel_t(hostname, vpid, kernel)];
+  state->profiled_function_name_and_ts[key] = fn_ts_t(name, ts);
+  state->function_name_to_dsd[hpt_function_name_t(hostname, vpid, vtid, name)] =
+    state->command_queue_to_device[hp_command_queue_t(hostname, vpid, command_queue)];
+}
```
TApplencourt, line 208: Key can be inlined.
and then all the type removed

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+                                           const char *event_class_name,
+                                           const char *hostname, int64_t vpid,
+                                           uint64_t vtid,
+                                           cl_command_queue command_queue) {
+  auto state = static_cast<data_t *>(usr_data);
+  hpt_t key{hostname, vpid, vtid};
+  auto name_str = strip_event_class_name_entry(event_class_name);
+  const char *name = name_str.c_str();
+  state->profiled_function_name_and_ts[key] = fn_ts_t(name, ts);
+  state->function_name_to_dsd[hpt_function_name_t(hostname, vpid, vtid, name)] =
```
TApplencourt, line 220: Remove type, and should we keep `name` as a string?

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,354 @@
...
+    state->profiled_function_name_and_ts[hpt_t(hostname, vpid, vtid)];
+
+  if (!state->event_to_function_name_and_ts.count(hp_event)){
+      state->event_to_function_name_and_ts[hp_event] =
+        tfn_ts_t(vtid, function_name, launch_ts);
+  } else {
+    const auto [start_ts, delta] = state->event_result_to_start_and_delta[hp_event];
+    const auto [device, subdevice] =
+      state->function_name_to_dsd[hpt_function_name_t(hostname, vpid, vtid, function_name)];
+
```
TApplencourt, line 240: Same remove type.

opencl/btx_clinterval_callbacks.hpp
bd4: ya can be moved to cpp as static inline.

opencl/btx_clinterval_callbacks.cpp
bd4: it works! Get confused about which cases it actually works for...

opencl/btx_clinterval_callbacks.cpp
bd4: I like idea of keeping name as string, then we don't have to worry about c ptr going dest going out of scope.

bd4: One issue with this, is all the compound types defined in xprof_utils, using char * for function name (e.g. hpt_function_name_t). But I guess if we limit using the name to only in the struct declaration, is less of an issue

opencl/btx_clinterval_callbacks.cpp
bd4: The map is set on `clGetDeviceIDs` and in `clCreateSubDevices`. I think the idea is that if you are using a top level device, you may never call these so the map would not be set and it would be 0, and would be a root device. This was copied from the old erb code.

bd4: Looking more at OpenCL API, cl_device_id is a pointer to a struct, so you can't easily skip the step of `clGetDeviceIDs` or `clCreateSubDevices`. I am unsure however if there is another API or something we are not covering, so hesitant to remove the if branch since it *might* break something and was in the old code.

TApplencourt: Agree, lets keep it for now! It's just a porting after all :) We can do refactoring in another pr

opencl/btx_clinterval_callbacks.cpp
```diff
@@ -0,0 +1,351 @@
...
+  btx_push_message_lttng_host(btx_handle, hostname, vpid, vtid, entry_ts,
+                              BACKEND_OPENCL, event_class_name_striped.c_str(),
+                              (ts - entry_ts), err);
+}
+
+void btx_initialize_component(void **usr_data) { *usr_data = new data_t; }
+
+void btx_finalize_component(void *usr_data) {
+  delete static_cast<data_t *>(usr_data);
```
TApplencourt, line 50: They can be static also

opencl/Makefile.am
```diff
@@ -237,41 +240,25 @@ xprof_utils.cpp: $(top_srcdir)/utils/xprof_utils.cpp
...
 	xprof_utils.hpp
 
 nodist_libCLInterval_la_SOURCES = \
-	babeltrace_clinterval.c \
-	babeltrace_clinterval.h \
-	babeltrace_clinterval_callbacks.h \
-	babeltrace_clinterval_dispatchers.c \
-	babeltrace_clinterval_callbacks.cpp \
-	babeltrace_clinterval_emitter.h \
-	tracer_opencl.h \
```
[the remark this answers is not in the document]
bd4: The `tracer_opencl.h` is not include directly by the new btx_clinterval* files, and if it is used indirectly, it should be reflected by the dependency chain. So I am pretty sure is not needed (and not sure if the tracer typedef's were used in old one either).

opencl/btx_clinterval_callbacks.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <iostream>
```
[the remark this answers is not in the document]
bd4: good catch, done

opencl/btx_clinterval_callbacks.cpp
bd4: done.

TApplencourt (approved):

PR #194, opened 2024-04-02
Fix visibility of profiling symbols.

Kerilk:
Those were getting exported, I don't think it was intended.


sbekele81:
Yes,  no need to be exported. Thanks.

sbekele81 (approved):

PR #201, opened 2024-04-15
Remove files

TApplencourt:
## Remove old test format

Before, we used a `yaml` format for the tests. It has proven to be a little tedious to update/maintain. The new tests use the output of `sink.text.pretty` and `babeltrace_thapy trace` (due to structs and stuff).
In the future, we should use the same output for everything :) 

## Remove all tests template

Some of the old tests where generated via a "template" common for all the backend. This is a "wrong good idea". In practice, all backends need some specialization.  

## Remove `xprof_utils.cpp`

It was used to handle pushing and creating a babeltrace2 message. Not required anymore

## Remove `gen_dispatcher`  and co

Now, all the dispatchers and plugins are handled via metababel. We can remove all those `gen` and `erb` in THAPI.

## Remove `interval.{c,h}.erb`. 

Not required anymore, now handled by metababel

## Remove libBabeltrace*

This is handled and tested by metababel. Old tests used them

## Remove dust/dummy trace

Dummy trace / dust was useful for testing our dispatcher. Thanks to metababel,  it's not required anymore.
Metababel will refuse to parse a `yaml` he doesn't know. And all the yaml is known to have been tested

## Remove lttng_dummy_trace.sh

It was not used. And we have, we have tracer.sh and integration test to test it now.


bd4 (commented):
Looks great! Why were some of the CI tests skipped?

TApplencourt:
Yes, the UI of GitHub is sometimes weird. 

All the CI tests were duplicated (one for the PR and one for my branch). So, to save some power, only one of those 2 similar tests has been run.

bd4 (dismissed):

bd4 (approved):

PR #207, opened 2024-04-22
Removed ddi table dumps as these are not stable between versions.

Kerilk:
While we wait for a more permanent solution...


Kerilk:
Permanent solution would need to use:
https://github.com/oneapi-src/level-zero-spec/blob/master/scripts/core/driver.yml
to deduce the different versions of the table and then dump the right one depending on the require level-zero version.

TApplencourt (approved):

TApplencourt:
From Intel

>MKL isn’t linked directly against the Level Zero library (it may not be present on the user’s system). So we do dynamically load the Level Zero library and use dlsym/GetProcAddress to find zeKernelCreate. So that may be the reason it’s not showing up in your trace.

ze/gen_ze.rb
```diff
@@ -220,12 +222,24 @@ def gen_struct_printer(namespace, types)
 
 $ze_commands.each { |c|
   normal_wrapper.call(c, :lttng_ust_ze, ze_struct_types)
+  puts <<EOF unless c.name.match(/zeGet.*ProcAddrTable/) unless c.name.match(/zeLoaderInit|zelLoaderDriverCheck|zelLoaderTracingLayerInit|zeLoaderGetTracingHandle/)
```
TApplencourt, line 225:
```suggestion
  puts <<EOF unless c.name.match(/zeGet.*ProcAddrTable/) && c.name.match(/zeLoaderInit|zelLoaderDriverCheck|zelLoaderTracingLayerInit|zeLoaderGetTracingHandle/)
```

TApplencourt (approved):

PR #208, opened 2024-04-22
Use zeKernelGetName

TApplencourt:


bd4 (changes requested):
LGTM, just need to remove debug print.

PR #213, opened 2024-04-29
ci: do plain make in dist-check job

bd4:


TApplencourt:
So this is passing too. Maybe it's more a `mpi_daemon` problem than anything else

bd4:
Ya I think there is an issue here with MPI detection but I think make check is a superset of make, so this PR should not be necessary.

PR #218, opened 2024-05-02
add iprof MPI daemon test

nscottnichols:
Adds CI test for MPI daemon. Not sure if the paths are correct for it to launch the test correctly.

### DIFF
--- a/integration_tests/iprof_mpi_daemon_test/test.sh
+++ b/integration_tests/iprof_mpi_daemon_test/test.sh
@@ -9,13 +9,27 @@ RT_SIGNAL_GLOBAL_BARRIER=$((SIGRTMIN + 1))
 RT_SIGNAL_LOCAL_BARRIER=$((SIGRTMIN + 2))
 RT_SIGNAL_FINISH=$((SIGRTMIN + 3))
 
+# Initialize a variable to track signal reception
+SIGNAL_RECEIVED="false"
+
 # Signal handler for capturing signals
 handle_signal() {
     echo "Received signal $1 from mpi_daemon"
+    if [ "$1" == "RT_SIGNAL_READY" ]; then
+        SIGNAL_RECEIVED="true"
+    fi
 }
 
 # Setup trap for ready signal sent from signal daemon
-trap 'handle_signal RT_SIGNAL_READY' $((SIGRTMIN + 0))
+trap 'handle_signal RT_SIGNAL_READY' $RT_SIGNAL_READY
+
+# Function to wait for RT_SIGNAL_READY
+wait_for_signal() {
+    SIGNAL_RECEIVED="false"
+    while [[ "$SIGNAL_RECEIVED" == "false" ]]; do
+        sleep 0.1  # Small sleep to prevent busy looping
+    done
+}
 
 # Get the PID of this script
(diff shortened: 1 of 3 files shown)


integration_tests/iprof_mpi_daemon_test/test.sh
```diff
@@ -0,0 +1,48 @@
...
+# Get the PID of this script
+PARENT_PID=$$
+
+# Start mpi_daemon in the background
+./mpi_daemon $PARENT_PID &
+DAEMON_PID=$!
+
+# Allow the daemon to initialize
+sleep 2
+
```
TApplencourt, line 23: Please don't do that. This this racy and nighmarish in a super-slow CI environment:

In ruby I do something like
```

  def lazy_exec(&block)
    return unless in_mpi_env?

    Signal.trap(RT_SIGNAL_READY) do
      return
    end
    block.call
    sleep
  end
  ```
  
  Aka: way that the daemon send a signal `RT_SIGNAL_READY` to the PARENT_PID.

File: integration_tests/iprof_mpi_daemon_test/test.sh [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+# Final synchronization after mpi_hello_world execution
+send_signal 2  # Again send RT_SIGNAL_LOCAL_BARRIER
+sleep 1
+
+send_signal 1  # Again send RT_SIGNAL_GLOBAL_BARRIER
+sleep 1
+
+# Signal to terminate the mpi_daemon
+send_signal 3  # RT_SIGNAL_FINISH is SIGRTMIN + 3
```
TApplencourt, line 47: may create some variable so we can use `RT_SIGNAL_LOCAL_BARRIER` ? Will be clearer IMO

integration_tests/iprof_mpi_daemon_test/test.sh
nscottnichols: Ah yes, will replace all sleeps with something similar.

integration_tests/iprof_mpi_daemon_test/test.sh
nscottnichols: Will do!

integration_tests/iprof_mpi_daemon_test/test.sh
nscottnichols: Added "wait for signal" function in 2f8d513.

integration_tests/iprof_mpi_daemon_test/test.sh
nscottnichols: Explicitly define signals in d81eba9.

integration_tests/iprof_mpi_daemon_test/test.sh
```diff
@@ -0,0 +1,48 @@
...
+#wait $DAEMON_PID  # Ensure daemon exits cleanly //FIXME this gives non-zero exit codes
```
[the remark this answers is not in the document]
nscottnichols: I think things were failing here because I missed capturing the last signal sometimes. It seems to work now with the new "wait for signal" function in d81eba9.

TApplencourt (approved):

PR #220, opened 2024-05-03
Add Ordinal Support

TApplencourt:
- Now metadata add `index` and `ordinal` of the command queue used
- All little more refactoring (better name, less data passed around)
- Add doc


bd4 (approved):
Looks good.

I wonder if some of the metadata could be structured in the future instead of just a string, but this seems like a nice bit of extra information and an improvement and a good place to start.

ze/tests/interval_profiling_multithread.bt_text_pretty
```diff
@@ -1,4 +1,4 @@
 lttng:host: { hostname = "testhost", vpid = 10, vtid = 2, ts = 1704110400000000000, backend = 1 }, { name = "zeCommandListAppendBarrier", dur = 20, err = false }
 lttng:host: { hostname = "testhost", vpid = 10, vtid = 1, ts = 1704110400000000000, backend = 1 }, { name = "zeCommandListAppendBarrier", dur = 20, err = false }
-lttng:device: { hostname = "testhost", vpid = 10, vtid = 2, ts = 1704110400000000000, backend = 1 }, { name = "zeCommandListAppendBarrier", dur = 20, did = 0, sdid = 0, err = false, metadata = "" }
-lttng:device: { hostname = "testhost", vpid = 10, vtid = 1, ts = 1704110400000000000, backend = 1 }, { name = "zeCommandListAppendBarrier", dur = 10, did = 0, sdid = 0, err = false, metadata = "" }
+lttng:device: { hostname = "testhost", vpid = 10, vtid = 2, ts = 1704110400000000010, backend = 1 }, { name = "zeCommandListAppendBarrier", dur = 20, did = 0, sdid = 0, err = false, metadata = "{ordinal: 0, index: 0}" }
```
TApplencourt, line 3: the time change, need to understand why, and if that make sense.

PR #226, opened 2024-05-09
Improve Traffic

TApplencourt:
- Add metadata for traffic 
- Add error checking for traffic (not sending traffic message if error)
    - For profiled commands, use event errors 
    - For non-profiled commands (zeMakeResidant and co), created a `_exit` callback

### DIFF
--- a/ze/btx_zematching_model.yaml
+++ b/ze/btx_zematching_model.yaml
@@ -152,6 +152,13 @@
       :type: structure
       :members:
       - :name: ^size$
+  - :set_id: eventMemory_without_hSignalEvent_exit
+    :domain: n = eventMemory_without_hSignalEvent_entry.map { |b| b.name.sub("_entry","_exit") }; exits.filter { |a| n.include?(a.name) }.to_set
+    :payload_field_class:
+      :type: structure
+      :members:
+      - :name: ^zeResult$
+
   - ## Rest
     :set_id: hSignalEvent_rest_entry
     :domain: hSignalEvent_entry - (hSignalEvent_hKernel_with_group_entry + hSignalEvent_hKernel_without_group_entry + hSignalEvent_eventMemory_entry)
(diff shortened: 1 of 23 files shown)


ze/btx_zematching_model.yaml
```diff
@@ -152,6 +152,13 @@
       :type: structure
       :members:
       - :name: ^size$
+  - :set_id: eventMemory_without_hSignalEvent_exit
+    :domain: n = eventMemory_without_hSignalEvent_entry.map { |b| b.name.sub("_entry","_exit") }; exits.filter { |a| n.include?(a.name) }.to_set
```
bd4, line 156: So this is creating a list of all the without hSignalEvent entries with "_exit" replaced, then searching through all exits for ones that actually exist? I guess not all _entry have corresponding exit? Took me a minute to map my head around. I think what is confusing about it, is the n being stored as a temporary, and what is actually "returned" is the second part.

bd4 (approved):
I wonder if the btx model could be made a little more clear, but overall I think it looks good, can always improve it more later if can think of a more elegant way to express it.

ze/btx_zematching_model.yaml
TApplencourt: > So this is creating a list of all the without hSignalEvent entries with "_exit" replaced, then searching through all exits for ones that actually exist?

Yes, just try to find the associated _exit to the `hSignalEvent_entry`. 

>  I guess not all _entry have corresponding exit? 

In generate case no, for our use-case in THAPI always.

> I think what is confusing about it, is the n being stored as a temporary, and what is actually "returned" is the second part.
- Typical ruby, where the last expression is what is returned. But yeah, if this is a useful partner we can create a metabaebl function call `get_opposite_domain` who far all `_entry`, `_exit` find their associated `_exit, _exit`

TApplencourt:
Using std::varient to add type safety to our metadata

bd4 (approved):
Looks good!

PR #227, opened 2024-05-13
Make only local master do energy profiling.

Kerilk:


TApplencourt:
Waiting for ZET change before merging

PR #228, opened 2024-05-14
fix sample

TApplencourt:
FIx:
 - remove old pretty printed
 - Use lttng ts to print correct start
 - Remove `name` in counter

Todo:
- Need to fix malloc / calloc  in ze_tracer_helper.c
- Also need may want to add `unit` to the protobuff (for watt, and %), https://perfetto.dev/docs/reference/trace-packet-proto#CounterDescriptor  (and use unit_multiplier if required)


Kerilk:
@TApplencourt pushed a commit that should address first TODO

TApplencourt:
This one is ready to merge I think

PR #229, opened 2024-05-14
fix min born

TApplencourt:
Using enqueue `_entry` for IMM and `submit`_entry for non IMM as min born for kernel launch

### DIFF
--- a/ze/btx_zeinterval_callbacks.cpp
+++ b/ze/btx_zeinterval_callbacks.cpp
@@ -278,7 +278,7 @@ static void zeCommandListCreateImmediate_exit_callback(void *btx_handle, void *u
     return;
 
   auto [hDevice, commandQueueDesc] = data->imm_tmp[{hostname, vpid, vtid}];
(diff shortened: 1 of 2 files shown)


ze/btx_zeinterval_callbacks.cpp
```diff
@@ -707,6 +709,7 @@ static void event_profiling_result_callback(void *btx_handle, void *usr_data, in
     }
     device_hash = hash_device(it0->second);
   }
+
```
Kerilk, line 712:
```suggestion
```

ze/btx_zeinterval_callbacks.cpp
```diff
@@ -699,6 +700,7 @@ static void event_profiling_result_callback(void *btx_handle, void *usr_data, in
   uint64_t delta = globalEnd - globalStart;
   uint64_t start = lltngMin;
   uintptr_t device_hash = 0;
+
```
Kerilk, line 703:
```suggestion
```

Kerilk (approved):

PR #233, opened 2024-05-31
add libthapi-ctl to start/stop collecting

bd4:
- for opencl backend, separates bookkeeping calls so they stay enabled even when client stops collection
- other backends will be fragile, making some API calls that do bookkeeping while collection is stopped may break things


bd4:
Basic functionality works, and opencl _should_ be robust. Needs better test, documentation, other backends to split out bookkeeping events.

Kerilk:
You need to add `liblttng-ctl-dev` to the apt install line.

TApplencourt:
Maybe you just have `libOpenCL.so.1` and not `libOpenCL.so`

PR #235, opened 2024-06-06
Fabric stat

sbekele81:
Sampling clean up & fabric sampling


xprof/btx_timeline.cpp
```diff
@@ -5,11 +5,16 @@
 #include <iomanip>  // set precision
 #include <iostream> // stdcout
 #include <map>
-#include <optional>
+//#include <optional>
```
TApplencourt, line 8: if `optional` not required, just remove it

sbekele81: Sure, will do that.

TApplencourt (changes requested):

PR #241, opened 2024-06-18
improvement in backend support

TApplencourt:
- Before `--backend XX` was broken. It always required to pass a level with it. Fixed now
- Before, it was impossible to ask the tally plugin to display only a specific backend. Now will respect the `--backends` passed.
- Print in reverse order of level (higher level first) make more sense.


bd4 (approved):
LGTM

PR #243, opened 2024-06-24
more control of FS

TApplencourt:
- Forbide user to make typo of FIX_FS
- Allow more control on where the lock is placed (aka use a different shared FS, this is the case on aurora where we should use flare and not HOME @colleeneb idea)


bd4 (approved):
LGTM

bd4 (approved):

bd4 (approved):

PR #246, opened 2024-07-01
sampling: add missing lttng-ust dep

bd4:


TApplencourt (approved):

TApplencourt:
Pinging brice for double checking

bd4:
The dependency is not obvious, but the generated `sampling.h` does do `#include <lttng/tracepoint.h>`, so fails to compile unless lttng is in include path. Easy to miss when lttng is installed in /usr/include.

Kerilk (approved):

PR #248, opened 2024-07-01
Fix LTTng array size types.

Kerilk:


bd4 (approved):
Fixes my build with gcc 12 and generated .tp looks correct now, thanks for the fix!

PR #249, opened 2024-07-01
More log and timeout in the CI

TApplencourt:
It was a transient error.  :( 
Adding more log, so at least we will know next time which bash command hanged


nscottnichols (approved):
LGTM

PR #252, opened 2024-07-09
Remove non events not used by interval

TApplencourt:
In the default mode, remove the events not used by interval.

- If not, they are passed around and finish to be written to disk with the `aggregated` events.
 - Then, during the replay, with it a `file` limits. 
 
 Of course, this is not a fix for the `file` limits but just a "cleaning" so that by default we just generate events that we use for the tally / interval.
 
 At term, we should allow user to specify events to enable/disable, and find a real solution for the file limits (smarted use of `--source` or  tree reduction)


TApplencourt:
Yeah, I was thinking we could do that, too. We cannot do for interval (as they are chained),  but we can do that for aggreg.  Or create a new plugins who drop everything who is not interval and put it as the end of the chain of interval. 

T the advantage that it may help timeline but if people want the timeline they will it the 2GB of protobuf before the number of file open :)

PR #254, opened 2024-07-11
Remove useless message in interval and aggreg

TApplencourt:
To remove the useless message we have two options:

- Either use the `--drop` in metababel at the creation of the component. This work well for Aggreg, but hard for Interval
- Modify the striper to be also a `filter`. This is was is done in the PR (relly on https://github.com/TApplencourt/metababel/pull/146) 

One current limitation is that the `stripper` can take only one`prefix` and not a list of prefix / regex. That may be a problem, as for example we may want to send  around the names of the devices (who are not part of the aggregate data class but maybe required for tally)

It's trivial to implement a `regex matching` on the stripper if required (a performance impact may be see for people doing the timeline, none for tally people)

It seems cleaner for me to have a `stripper / filter` to modify the behavior of intermediate components, as it's more general than particular components dropping unknown messages.  But the "drop message" is arguably a more straightforward. It only works for aggregates, but it is simple. And at the end `aggregate` is really what matter.

So 🤷🏽 

I will let you decide, it was not too painful to implement as a concept, so can revert back.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -77,7 +77,7 @@ PKG_CHECK_MODULES([PROTOBUF], [protobuf >= 3.0])
 AX_RUBY_EXTENSION([cast-to-yaml], [yes])
 AX_RUBY_EXTENSION([nokogiri], [yes])
 AX_RUBY_EXTENSION([babeltrace2], [yes])
-AX_RUBY_EXTENSION([metababel >= 1.1.1], [yes])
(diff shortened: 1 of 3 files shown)


TApplencourt:
PS: Of course the CI failed due to the metababel not yet released (https://github.com/TApplencourt/metababel/pull/146). When we decide the drop / filter, I can release a new metababel version.

PR #270, opened 2024-07-24
Simplify tests

TApplencourt:
This will make tests for the backends easier.

We can add new `.bats` call `backend tests` that run simple tests. (mpi_hello world, and OpenCL hello world -- does the CI can build opencl kernel? if not clinfo) ).

For stuff that is nonaccessible in the CI, we need another option.

I may add a tests to run a MPI apps  after this is merged (but I'm a little afraid that mpi session will be broken...).


.github/workflows/presubmit.yml
```diff
@@ -134,22 +134,14 @@ jobs:
           name: thapi-bin
           path: thapi.tar
 
-  integration-tests:
+  Integration-tests:
```
Kerilk, line 137:
```suggestion
  integration-tests:
```

TApplencourt: All other step have start capitlaized

TApplencourt: Oh no but they show in the CI capilized... humm

TApplencourt: Wanted to avoid the lower-case in `https://github.com/argonne-lcf/THAPI/actions/runs/10080775871/job/27871485893` "Load Babeltrace" but then "integration test"

TApplencourt: Ok fixed in 3edff95

bd4 (dismissed):
Nice, looks good!

integration_tests/parrallel_execution.bats
Kerilk, line 1: parrallel => parallel

Kerilk: Pretty sure this file should be referenced somewhere, but I can't find where it is.

TApplencourt: `bats ./integration_tests` in the CI will run all the `*.bats` inside the folder.

TApplencourt: fixed with acecaae

Kerilk (approved):

PR #272, opened 2024-07-26
Prevent wrapping unsupported version of CUDA functions.

Kerilk:
To discuss.

### DIFF
--- a/cuda/cuda_model.rb
+++ b/cuda/cuda_model.rb
@@ -299,7 +299,9 @@ register_proc_callbacks = lambda { |method|
   str = <<EOF
   const int pt_condition = #{pt_condition};
   const int normal_condition = #{normal_condition};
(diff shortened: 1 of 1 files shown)


TApplencourt:
LGTM!
No sure if we need to escape the `\` or `\n` or not

TApplencourt:
The `\` doesn't show up
```
THAPI: CUDA version 12020 is unsupported, could not wrap cuMemPrefetchAsync symbol
```
So I guess we are good.

One of the problem, is that it may break the analysis (for example if don't wrap cuModuleLoad or whatever)

bd4:
Is this for when the runtime lib version is newer than the version used at build time?

TApplencourt:
Exactly! IE when we didn't update the header in THAPI

TApplencourt (approved):

PR #274, opened 2024-07-30
add missing entry

TApplencourt:
Some memcpy exit didn't get their associated `entry`. We where missing some traffic entry .

I don;t know if I compute the `size` of the data transfert correctly


bd4 (approved):
Looks good!

PR #280, opened 2024-08-28
Fix  OpenCL Start

TApplencourt:
Not really sure what was the bug, but the new version seem to work and the code is clearer. 

Fix @colleeneb  bug with the OpenCL Timeline


TApplencourt:
The CI is just due one more freasking time to the MPI. I have no clue, those bug have been fixed already 10 time...

Kerilk (approved):

PR #281, opened 2024-08-28
[draft] fix ci

TApplencourt:


TApplencourt:
We use on one node, so in theory should work. Need to understand more this stupid transient error (https://github.com/argonne-lcf/THAPI/actions/runs/10603787267/job/29389161346)

PR #282, opened 2024-08-30
Fix typos

TApplencourt:
@Kerilk found a typo.

It doesn't impact the pass rate, as our convertor from the log doesn't check for argument names. Just take the value, and call `push_message_$function_name($values,...)`


Kerilk (approved):

bd4:
Good catch!

PR #284, opened 2024-09-03
Fix Race Condition numero2323

TApplencourt:
To avoid race conditions, `SIGNAL_RECEIVED` needs to be set  before spawning or signaling the daemon


Kerilk (approved):
LGTM.

PR #324, opened 2025-02-13
add c++17 check

TApplencourt:
Not sure if it will crash or need to do a 
```
if test "$HAVE_CXX17" = "yes"; then
  AC_MSG_RESULT([C++17 support is enabled])
else
  AC_MSG_ERROR([C++17 support is required but not found])
fi
```
If somebody will a old gcc can try will be nice

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -81,6 +81,8 @@ AX_RUBY_EXTENSION([nokogiri], [yes])
 AX_RUBY_EXTENSION([babeltrace2], [yes])
 AX_RUBY_EXTENSION([metababel >= 1.1.2], [yes])
 
+AX_CXX_COMPILE_STDCXX([17], [noext], [mandatory])
+
 # Checks for header files.
 AC_CHECK_HEADERS([inttypes.h stddef.h stdint.h stdlib.h string.h unistd.h])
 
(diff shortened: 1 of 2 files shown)


Kerilk:
Isn't the default compiler for sunspot or JLSE a gcc 7.5?

rpereira-dev (changes requested):
Forwarding discussion we had with @TApplencourt 
- The patch is correct
- It does not fix its motivational issue
The original issue was `spack install thapi` failing with GCC 7.5
It fails building Magic Enum - see compiler compatibility for details: https://github.com/Neargye/magic_enum?tab=readme-ov-file#compiler-compatibility

A few more bits on C++17 since then - I tried to build THAPI with GCC
- 5.5.0 (do not support C++17) and 7.5.0 (supports C++17) -> build fails on Magic Enum
- 9.2.0 -> build passes

As discussed with Thomas, we should check for Magic Enum rather than C++17 compatibility

TApplencourt:
>5.5.0 (do not support C++17)

Does it at least now fail on the `configure`? 🤞🏽

TApplencourt:
What should we do with this path? Maybe good enough to require c++17. I may start to use c++17 feature anyway

TApplencourt:
Merged because, not perfect, but better than what we have now

PR #328, opened 2025-03-06
Add tmate support for debugging GitHub CI failures

thilinarmtb:
@TApplencourt : Do you think this is worth merging for debugging the GitHub CI?

This is similar to what we did [here](https://github.com/argonne-lcf/sycltrain)

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -1,5 +1,12 @@
 name: Presubmit
 on: [push, pull_request]
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
+        type: boolean
+        description: 'Run the build with tmate debugging enabled'
+        required: false
+        default: true
 
 env:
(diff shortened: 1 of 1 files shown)


TApplencourt:
Maybe just for the `Integration tests`?
For the other, we generate a `log` as an archive. So, it should be enough to debug. 

What do you think @Kerilk ? 

Having a way to `ssh` for the integration tests seems nice. I'm just afraid to always wait 1800s or whatever for each failure. (oh looks like we have ` timeout-minutes: 15` and th `workflow_dispatch`)

thilinarmtb:
Sounds good. I can just keep this for `integration tests`. I will wait for @Kerilk's input.

Kerilk:
Nice to have around, maybe we can keep it here and cherry pick the commit when we need to debug stuff.

TApplencourt:
Let keep the branch, and close the PR for now (to tired to find a better solution for now)

PR #334, opened 2025-03-21
Refactor ci

TApplencourt:
Actions are really nice!

### DIFF
--- a/.github/actions/setup-thapi/action.yml
+++ b/.github/actions/setup-thapi/action.yml
@@ -2,6 +2,54 @@ name: 'Setup Thapi Environment'
 runs:
   using: "composite"
   steps:
+    - uses: actions/cache@v4
+      id: efficios_dep
+      with:
+        path: ~/efficios_dep/
+        key: ${{ runner.os }}-build-cache-efficios_dep
+    - name: Build dependencies
+      if: steps.efficios_dep.outputs.cache-hit != 'true'
+      shell: bash
+      run: |
+        # Dependency for the install
+        sudo apt update
+        sudo apt install -y gcc g++ elfutils libelf-dev libpopt-dev libdw-dev libglib2.0-dev libnuma-dev liburcu-dev
+        # So dependency can find themself
+        export "PKG_CONFIG_PATH=$HOME/efficios_dep/lib/pkgconfig:$PKG_CONFIG_PATH"
+
+        # lttng-ust
+        git clone https://github.com/lttng/lttng-ust
+        cd lttng-ust
+        git checkout 4f8afc535e77070f1ef00434674f0417c6f9ef69
+        ./bootstrap
+        ./configure --disable-man-pages --prefix=$HOME/efficios_dep/
+        make -j$(nproc)
+        make install
+        cd ..
+ 
+        # lttng-tools
+        git clone -b anl-ms3 git://git.efficios.com/deliverable/lttng-tools.git
+        cd lttng-tools
+        ./bootstrap
+        ./configure --disable-man-pages --disable-bin-lttng-crash --prefix=$HOME/efficios_dep
+        make -j$(nproc)
+        make install
+        #Todo, this will need to be added in the spack repo as a patch
+        echo "#!/usr/bin/env python"| cat - dirwatch.py > $HOME/efficios_dep/bin/dirwatch.py
+        chmod 755 $HOME/efficios_dep/bin/dirwatch.py
+        cd ..
+
+        # babeltrace
+        git clone -b anl-ms3 git://git.efficios.com/deliverable/babeltrace.git
+        cd babeltrace
+        wget https://raw.githubusercontent.com/argonne-lcf/THAPI/53262fcaaaf45d7d475884d7e63b69abe47e41d6/.github/workflows/str_nullptr.patch
+        patch -p1 < str_nullptr.patch
+        wget https://raw.githubusercontent.com/argonne-lcf/THAPI/4418916620496fd66cde0b3d5e241bed0a4c18a3/.github/workflows/bt_makefile.patch
+        patch -p1 < bt_makefile.patch
+        ./bootstrap
+        ./configure --disable-man-pages --prefix=$HOME/efficios_dep
+        make -j$(nproc)
+        make install
     - name: Set Efficios environment variables
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


.github/actions/setup-thapi/action.yml
```diff
@@ -0,0 +1,73 @@
+name: 'Setup Thapi Environment'
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v4
```
Kerilk, line 5: Why checking out the repo here? I don't think you use it, and need to recheck different versions later in some actions.

TApplencourt: Good point!

TApplencourt, line 5:
```suggestion
```

TApplencourt:
Ok ready to merge I think.

(PS: the `compare` step "work" when th PR is not open, not sure what it's doing but it's not failing at least...)

.github/workflows/presubmit.yml
```diff
@@ -243,80 +125,47 @@ jobs:
...
-      - run: ./autogen.sh
-      - run: ../configure
-        working-directory: build
-      - run: make dist
-        working-directory: build
+          ./autogen.sh
+          mkdir -p build
+          cd build
+          ../configure
+          make dist
```
Kerilk, line 159: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
```diff
@@ -243,80 +125,47 @@ jobs:
...
-      - run: ./autogen.sh
-      - run: ../configure
-        working-directory: build
-      - run: make distcheck -j
-        working-directory: build
+          ./autogen.sh
+          mkdir -p build
+          cd build
+          ../configure
+          make -j distcheck
```
Kerilk, line 141: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
```diff
@@ -182,55 +87,32 @@ jobs:
...
-          echo "LD_LIBRARY_PATH=$HOME/efficios_dep/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
-          echo "LIBRARY_PATH=$HOME/efficios_dep/lib:$LIBRARY_PATH" >> $GITHUB_ENV
-      - run: ./autogen.sh
-      - run: ./configure
-      - run: make -j
-      - run: make check
+          ./autogen.sh
+          ./configure
+          make -j
+          make check
```
Kerilk, line 115: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
```diff
@@ -140,28 +59,14 @@ jobs:
...
-      - run: ./autogen.sh
-      - run: ../configure --prefix=`pwd`/ici
-        working-directory: build
-      - run: make -j install
-        working-directory: build
+          ./autogen.sh
+          mkdir -p build
+          cd build
+          ../configure --prefix=`pwd`/ici
+          make -j install
```
Kerilk, line 69: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
```diff
@@ -92,33 +26,18 @@ jobs:
...
-      - run: make check
-        working-directory: build
+      - name: Make check
+        run: |
+          ./autogen.sh
+          mkdir -p build
+          cd build
+          ../configure
+          CC=gcc make -j
+          make check
```
Kerilk, line 40: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
Kerilk, line 115: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
Kerilk, line 159: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
```diff
@@ -243,80 +125,47 @@ jobs:
...
-          cd `find . -type d -name "thapi-*"`
+          cd $(find . -type d -name "thapi-*")
           mkdir -p build
           cd build
           ../configure
-        working-directory: build
-      - name: Check
-        run: |
-          cd `find . -type d -name "thapi-*"`
-          cd build
```
Kerilk, line 319: I don't like these, this goes contradictory to the philosophy of github actions and increasing the diff for no good reason. I think that if we want to do that we should do so in a different PR.

.github/workflows/presubmit.yml
TApplencourt: I can revert and do another PR for sure, but YAML is just a stupid format for any kind of programming or control flow. 

We should just use bash at maximum possible aka, limiting the number of `run` and magic `work-directory`, ` and stuff.

Kerilk: I don't agree, we're loosing meta information by doing this. Let's chat about it.

.github/actions/setup-thapi/action.yml
```diff
@@ -0,0 +1,79 @@
...
+      shell: bash
+      run: sudo gem install cast-to-yaml nokogiri babeltrace2 opencl_ruby_ffi metababel
+
+    - uses: actions/cache@v4
+      id: efficios_dep
+      with:
+        path: ~/efficios_dep/
+        key: ${{ runner.os }}-build-cache-efficios_dep
+
+    - name: Install Efficios PT Dependencies
```
Kerilk, line 21:
```suggestion
    - name: Install Efficios APT Dependencies
```

Kerilk (approved):
Looking great!

PR #337, opened 2025-03-25
Fix ffi fixed sized integers.

Kerilk:
After this patch, we should not have any :uintxx_t inside the generated ruby libraries, but :uintxx

Will need to fix babeltrace2 bindings as well

### DIFF
--- a/utils/gen_library_base.rb
+++ b/utils/gen_library_base.rb
@@ -14,6 +14,8 @@ def to_ffi_name(name, default = true)
     return ':uint64'
   when 'size_t'
     return ':size_t'
+  when /^(u{0-1}int\d+)_t$/
+    return ":#{$1}"
   when '_Bool'
     return ':bool'
   end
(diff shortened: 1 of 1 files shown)


TApplencourt:
Then the `[Check same generated files]` is broken, it should have changed something no?

Kerilk:
Yeah maybe I fumbled the regexp, in the cab right now...

utils/gen_library_base.rb
```diff
@@ -14,6 +14,8 @@ def to_ffi_name(name, default = true)
     return ':uint64'
   when 'size_t'
     return ':size_t'
+  when /^(u{0-1}int\d+)_t$/
```
Kerilk, line 17:
```suggestion
  when /^(u?int\d+)_t$/
```

TApplencourt:
So before it was `:uint32-_t` and now it's `uint32`? Or I screw up the `new` / `ref` order

Kerilk:
:unt32_t becomes :uint32

TApplencourt:
Ok perfect, so the diff look good.

Kerilk:
If you could check there are no leftovers then we can merge

PR #364, opened 2025-06-25
[DRAFT] Sampling daemon loop

nscottnichols:
Fix mem leak and busy wait in sampling daemon

### DIFF
--- a/ze/sampling_daemon.c
+++ b/ze/sampling_daemon.c
@@ -790,10 +790,12 @@ int main(int argc, char **argv) {
   if (kill(parent_pid, RT_SIGNAL_READY) != 0) {
     _ERROR_MSG("Failed to send READY signal to parent");
   }
-  // Process_sampling loop until SIG_SAMPLING_FINISH signal
+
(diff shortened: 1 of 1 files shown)


nscottnichols:
This fixes the memory leak from process_sampling() registering the sampling function infinitely in the while loop. Also removes the busy wait by pausing and waiting for a signal in the while(). The sampling function is registered once and handled by the spawned sampling loop in [sampling/thapi_sampling.c#L117](https://github.com/argonne-lcf/THAPI/blob/fcbedb49bb1a94c44c79d72c364ea90618617621/sampling/thapi_sampling.c#L117).

We need to wait for @TApplencourt to fix the sampling spawner to include the appropriate environment variables for this. Otherwise the control variable `LTTNG_UST_SAMPLING` needs to be manually set.

TApplencourt:
Please rebase, and put back THAPI_SAMPLING :)

TApplencourt:
As usual github doesn't like the rebase, but LGTM if that work : ) -- Can you try to add in the `interagration testing` a simple call to `--sampling` if you don't mind? Will not do much but at least give me a little more confidence...

PR #365, opened 2025-06-25
Fix sampling daemon not getting the env

TApplencourt:
Before, sampling daemon didn't get the ENV variable set by xprof.

- That mean it was broken, as the LD_LIBRARY_PATH was not set, so was not able to find the tracing.so lib. It was working just because the spack module was setting the LD_LIBRARY_PATH
- @nscottnichols PR will put back SAMPLING env, right now it make it hang if we put it due to bugs in the ze sampling code.
- In the future, we can just not set `LD_LIBRARY_PATH` for the sampling daemon, this should simplify the ze sampling daemon a lot by allowing him to call direcly ZE symbol and not dlopen the "original" lib

### DIFF
--- a/xprof/xprof.rb.in
+++ b/xprof/xprof.rb.in
@@ -384,13 +384,13 @@ class SamplingDaemon < SpawnDaemon
     lazy_exec(message, sampling?, &block)
   end
 
-  def initialize
+  def initialize(h = {})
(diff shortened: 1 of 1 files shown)


TApplencourt:
Prerequisite of #364

Kerilk (approved):
LGTM

PR #367, opened 2025-06-27
Add sampling intergration tests

TApplencourt:


TApplencourt:
Ugly, but at least we have a CI. Should give us confidence on future refactoring.

PR #368, opened 2025-07-01
Use C++ compatible C headers

thilinarmtb:


TApplencourt:
`<assert.h>` is not C++?  ;'0

thilinarmtb:
I think it is. I mainly made these changes to be more inline with other places in THAPI where C
headers are include as `<cname>` rather than `<name.h>`. These were kind of the only exceptions.

TApplencourt:
You are right, we are in c++ let use c++ header :)

TApplencourt (approved):

PR #369, opened 2025-07-01
Introduce `thapi_sampling_daemon` 

TApplencourt:
So the diff is bigger than expected because I'm stupid but, in short the change are small.

We now have one `thapi_sampling_daemon`, who can open some `libThapiSamplingPluging.so`. `libThapiSamplingPluging` just need to expose a `thapi_register_sampling_plugin`  function, and inside call `thapi_register_sampling`.

In short we:
- Introduce `libThapiSampling.so`
     - Expose only 3 symbols, the `register`, `init`, and `sampling_loop`
- Introduce `thapi_sampling_register.h` header, exposing only `register`, the only header than sampling plugin need
- Introduce `thapi_sampling.h` for the `thapi_sampling_daemon` so `sampling_loop` and `init`.

This will help nathan and is new CXI plugin sample.

In the futur we can maybe clean `thapi_sampling_daemon` (we don't need pthread protection anymore) and the `ze_sampling_pluging` (no need to dl_open anymore, we can just call directly the pointers)

### DIFF
--- a/sampling/thapi_sampling.c
+++ b/sampling/thapi_sampling.c
@@ -7,13 +7,20 @@
 #include "thapi_sampling.h"
 #include "sampling.h"
 #include "utarray.h"
+#include <signal.h>
+#include <time.h>
+
+#define RT_SIGNAL_READY (SIGRTMIN)
+#define RT_SIGNAL_FINISH (SIGRTMIN + 3)
 
 struct sampling_entry {
-  void (*pfn)(void);
+  void (*pfn_run)(void);
   struct timespec interval;
+  void (*pfn_final)(void);
   struct timespec next;
 };
 
+typedef void (*plugin_init_func)();
 
 static pthread_mutex_t thapi_sampling_mutex = PTHREAD_MUTEX_INITIALIZER;
 static UT_array *thapi_sampling_events = NULL;
@@ -21,18 +28,26 @@ static UT_array *thapi_sampling_events = NULL;
 static pthread_once_t thapi_init_once = PTHREAD_ONCE_INIT;
 static volatile int thapi_sampling_finished = 0;
 static volatile int thapi_sampling_initialized = 0;
-static pthread_t thapi_sampling_thread;
+
+static void signal_handler_finish(int signum) {
+  printf("signal_handler_finish\n");
+  if (signum == RT_SIGNAL_FINISH) {
+    thapi_sampling_finished = 1;
+  }
+}
 
 static void __attribute__((destructor))
 thapi_sampling_cleanup() {
   if (!thapi_sampling_initialized)
     return;
   thapi_sampling_finished = 1;
-  pthread_join(thapi_sampling_thread, NULL);
   pthread_mutex_lock(&thapi_sampling_mutex);
   struct sampling_entry **entry = NULL;
-  while ((entry = (struct sampling_entry **)utarray_next(thapi_sampling_events, entry)))
+  while ((entry = (struct sampling_entry **)utarray_next(thapi_sampling_events, entry))) {
+    if ( (*entry)->pfn_final )
+        (*entry)->pfn_final();
     free(*entry);
+  }
   utarray_free(thapi_sampling_events);
   pthread_mutex_unlock(&thapi_sampling_mutex);
 }
@@ -66,7 +81,7 @@ static inline void time_add(struct timespec *dest, const struct timespec *t, con
   }
 }
 
-void * thapi_sampling_loop(void *args) {
+static void * thapi_sampling_loop(void *args) {
   (void)args;
   while(!thapi_sampling_finished) {
     struct timespec now;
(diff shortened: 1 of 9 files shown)


sampling/thapi_sampling.c
```diff
@@ -1,43 +1,41 @@
...
   struct timespec next;
 };
 
-
 static pthread_mutex_t thapi_sampling_mutex = PTHREAD_MUTEX_INITIALIZER;
 static UT_array *thapi_sampling_events = NULL;
 
 static pthread_once_t thapi_init_once = PTHREAD_ONCE_INIT;
-static volatile int thapi_sampling_finished = 0;
+volatile int thapi_sampling_finished = 0;
```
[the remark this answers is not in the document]
TApplencourt: Yeah, I replaced it via a function. Cleaner.

But the reason was that because it was exported in the header, we need `extern` : 
- and we cannot have `extern + static`.
- `extern` cannot be initlized
Then the trick of `header` + definition once (without the static)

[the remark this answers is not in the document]
TApplencourt: See `thapi_sampling_stop` function

nscottnichols (commented):
LGTM, will let @Kerilk do approval

sampling/thapi_sampling.h
Kerilk, line 1: Rollback changes to this file.

sampling/thapi_sampling.c
Kerilk, line 1: Rollback changes to this file

sampling/thapi_sampling.h
TApplencourt: It's here `sampling/thapi_sampling_register.h`

sampling/thapi_sampling.c
TApplencourt: I can put back `thapi_unregister_sampling`

TApplencourt:
>The reason for the register/unregister is that you may want to start and stop sampling programmatically.

I will add it back. Even if I don't see how can we using in the current algorithm. The pluging just call expose one function, who will be called once, who will call registration. I suppose the plugin so can spawn fork in the `plugin_registration` and 20min later `unregister` this is true.

- We need to add a `finalization` callback, so user pluging can register some finalization step for their own plugin (for example closing the handle)
- We need to a `thapi_loop_stop` function so we don't expose the `extern volatile` to the daemon_binary

--- 
- Why do we need to spawn thread anymore? Now iprof force the daemon to be run only by one process.
- Same for the pthread mutex

But I agree unreleated to the PR, we can do in another one.

TApplencourt:
- The CI is passing \o/.
- Using the "new-old" sampling daemon.
- Add an example "heartbeat" plugin

Kerilk:
For code generation purposes, I think it is simpler to have sampling plugins with their respective backends. For a distribution perspective I think the current scheme works. I have questions about the capitalization though, here is the list of files we have currently:
```
lib/thapi/sampling
lib/thapi/sampling/libThapiSampling.la
lib/thapi/sampling/libheartbeatsampling.so
lib/thapi/sampling/libThapiSampling.so.1
lib/thapi/sampling/libThapiSampling.so
lib/thapi/sampling/libheartbeatsampling.la
lib/thapi/sampling/libThapiSampling.so.1.0.0
[...]
here/lib/thapi/bt2/libCUDAInterval.so
here/lib/thapi/bt2/libOMPInterval.so
here/lib/thapi/bt2/libZEInterval.so
here/lib/thapi/bt2/libXStripper.so
here/lib/thapi/bt2/libXTimeline.so
here/lib/thapi/bt2/libCLInterval.la
here/lib/thapi/bt2/libXTally.so
here/lib/thapi/bt2/libXTimeline.la
here/lib/thapi/bt2/libOMPInterval.la
here/lib/thapi/bt2/libHIPInterval.so
here/lib/thapi/bt2/libXTally.la
here/lib/thapi/bt2/libMPIInterval.la
here/lib/thapi/bt2/libXAggreg.la
here/lib/thapi/bt2/libZEInterval.la
here/lib/thapi/bt2/libMPIInterval.so
here/lib/thapi/bt2/libXAggreg.so
here/lib/thapi/bt2/libCLInterval.so
here/lib/thapi/bt2/libXStripper.la
here/lib/thapi/bt2/libHIPInterval.la
here/lib/thapi/bt2/libCUDAInterval.la
[...]
here/lib/thapi/ze/libze_loader.la
here/lib/thapi/ze/libze_loader.so
here/lib/thapi/ze/libzesampling.so
here/lib/thapi/ze/libze_loader.so.1
here/lib/thapi/ze/libze_loader.so.1.0.0
here/lib/thapi/ze/libzesampling.la
```
I think we should be consistent about naming our libraries, and I like the capitalized version better to be honnest.
So I propose renaming:
 * `libzesampling.so` => `libZESampling.so`
 * `libheartbeatsampling.so` => `libHeartbeatSampling.so`

TApplencourt:
Consistency always good!  

 I guess I just got triped by `libze_loader`. And if I remenber correctly inside the different Makefile we are not super consistance neither.  (also the opencl / cl, I guess we should fix that one day too)

I will try to make a `README_NAMING_CONVENSION` or something in a separate PR and will fix the Saming into this one

Kerilk:
We can't change the ones which are not ours like `libze_loader.so`. I double checked the rest, we are pretty consistent, I think we already went over them once,

Kerilk:
I forgot about non-installed libraries. Agreed, Camel case is good and we should use it consistently when possible.

Kerilk (approved):
LGTM

PR #384, opened 2025-07-11
Use default values for test env vars

thilinarmtb:
Also, set all the env vars in a common file. This will make it easier to run tests manually
and through spack in the future.

### DIFF
--- a/integration_tests/test_env_vars.bash
+++ b/integration_tests/test_env_vars.bash
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 export THAPI_HOME=${THAPI_HOME:-${PWD}}
-export THAPI_BIN_DIR=${THAPI_BIN_DIR:-${PWD}/build/ici/bin}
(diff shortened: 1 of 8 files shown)


integration_tests/test_env_vars.bash
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+export THAPI_HOME=${THAPI_HOME:-${PWD}}
+export THAPI_INSTALL_DIR=${THAPI_INSTALL_DIR:-${PWD}/build/ici}
```
TApplencourt, line 4: - Why we need `THAPI_INSTALL_DIR` ?

It seem to be used onky for `THAPI_BIN_DIR`, and was the only variable set before.
So I suggest to remove it.

thilinarmtb: You are correct -- we don't seem to use it anywhere.

TApplencourt (approved):

PR #397, opened 2025-08-29
use anl-v2.14.0 branch of lttng-tools for archive

thilinarmtb:


thilinarmtb:
Yeah -- I thought I had to wait till you approved. Was it because there was a cache hit?

TApplencourt:
Let me run it manually just to double check : )

TApplencourt:
Oh it's because it's from your fork, to avoid "side effect" attak

thilinarmtb:
seems like it is failing -- I will ping when it passes.

PR #398, opened 2025-08-29
Only generate btx

TApplencourt:


TApplencourt:
OpenCL is special enough that i will do it in another PR, so we can remove `/utils/babeltrace_to_metababel_model.rb`. I'm too tired : )

nscottnichols (approved):
LGTM!

PR #400, opened 2025-09-05
Move last backend (opencl) to natively generated new btx model

TApplencourt:


TApplencourt:
The only thing I'm not particularly sure is the `    if cast_type.include?('*') && parsed_field[:field_class].include?(:element_field_class) && field['lttng'] != 'ctf_array'` 

I guess I case of ctf_array, the type is always `size_t *` for opencl, and before we never casted back this to `size_t` for some particular reason

PR #413, opened 2025-09-24
Update README.md

TApplencourt:
I think we should update / modernize the Readme. 

I added what I think was usefull (for example, colleen didn't knew autotool well)


TApplencourt:
I expect this to be a collab PR please feel to push into the main. 
For example more picture, more doc, more refactoring (iprof before tracer? Is the tracer that used? Description of the table and more option, this kind of thing)

TApplencourt:
Oh good idea, the `please cite us` thingy with the bibtex!

PR #414, opened 2025-09-24
[CUDA] Use RTLD_DEFAULT in dlsym to check for self open

nscottnichols:
Fixes the symbol redefinition issue I was seeing by using RTLD_DEFAULT tells dlsym to fetch the same cudaSetDevice you’d get from a normal call at runtime.

My understanding is the first `dlsym(handle, "cudaSetDevice")` gets the address `cudaSetDevice` resolves to within the symbol scope and then by using `RTLD_DEFAULT` in the `dlsym(RTLD_DEFAULT, "cudaSetDevice")` call gives the address for the `cudaSetDevice` called within that translation unit. And then, you can just compare the two pointers to see if you are opening yourself.

### DIFF
--- a/backends/cuda/tracer_cudart_helpers.include.c
+++ b/backends/cuda/tracer_cudart_helpers.include.c
@@ -14,9 +14,10 @@ static void _load_tracer(void) {
       handle = dlopen("libcudart.so", RTLD_LAZY | RTLD_LOCAL);
   if (handle) {
     void* ptr = dlsym(handle, "cudaSetDevice");
-    if (ptr == (void*)&cudaSetDevice) { //opening oneself
-      dlclose(handle);
-      handle = NULL;
+    void* self = dlsym(RTLD_DEFAULT, "cudaSetDevice");
+    if (ptr == self) {  /* opening oneself */
+        dlclose(handle);
+        handle = NULL;
     }
   }
 
--- a/backends/cuda/gen_cudart.rb
+++ b/backends/cuda/gen_cudart.rb
@@ -1,6 +1,8 @@
 require_relative 'cudart_model'
 
 puts <<EOF
+#define _GNU_SOURCE
+#include <dlfcn.h>
 #define __CUDA_API_VERSION_INTERNAL 1
 #include <cuda_runtime_api.h>
 #include <pthread.h>


backends/cuda/tracer_cudart_helpers.include.c
```diff
@@ -14,9 +14,10 @@ static void _load_tracer(void) {
       handle = dlopen("libcudart.so", RTLD_LAZY | RTLD_LOCAL);
   if (handle) {
     void* ptr = dlsym(handle, "cudaSetDevice");
-    if (ptr == (void*)&cudaSetDevice) { //opening oneself
-      dlclose(handle);
-      handle = NULL;
+    void* self = dlsym(RTLD_DEFAULT, "cudaSetDevice");
+    if (ptr == self) {  /* opening oneself */
+        dlclose(handle);
+        handle = NULL;
```
Kerilk, line 20: Have you tried opening yourself by pointing `LTTNG_UST_CUDART_LIBCUDART` to the tracer, and verified it tripped the check?
If you have then I am fine approving it, I agree that it should be good enough. Alternative would be to define a hidden alias before the symbol gets deleted and the use that to check:
```c
static cudaError_t cudaSetDevice_hid(int device) __attribute__ ((alias ("cudaSetDevice")));
```

nscottnichols: Oh, good point! No I didn't test that, will check now.

nscottnichols: > Is this PR have 3 fix?
GNU_SOURCE for better parsing, dlfcn for dlsym, and then the RTLD_DEFAULT thingy?

`GNU_SOURCE` is needed to get `RTLD_DEFAULT` extension from `dlfcn.h`.

The include was because I wasn't sure where `dlfcn.h` was coming in (somewhere in the CUDA headers is my guess) and wanted to ensure `GNU_SOURCE` wasn't unset before it was included.

nscottnichols:
> Is this PR have 3 fix?
GNU_SOURCE for better parsing, dlfcn for dlsym, and then the RTLD_DEFAULT thingy?

`GNU_SOURCE` is needed to get `RTLD_DEFAULT` extension from `dlfcn.h`.

The include was because I wasn't sure where `dlfcn.h` was coming in (somewhere in the CUDA headers is my guess) and wanted to ensure `GNU_SOURCE` wasn't unset before it was included.

backends/cuda/tracer_cudart_helpers.include.c
Kerilk: Yes, thanks!

Kerilk: I assume this:
```
Don't want to open self, closing...
```
Is a printf you've added yourself correct? otherwise I am unsure where it is coming from.

Kerilk (approved):

backends/cuda/tracer_cudart_helpers.include.c
nscottnichols: Yes, exactly, I should have mentioned that. Also the "`in _load_tracer`".

PR #418, opened 2025-09-25
H2yaml versioning

TApplencourt:
So before i was ugly, and just pip install h2yaml last upstream version...  I did put versioning check and passing to the new h2yaml API (the 4th time I changed the API for the CLI, I think this time is the good one)


nscottnichols (approved):
LGTM!

PR #421, opened 2025-09-30
Fix gen_probe_base for function without parameter returning something

TApplencourt:
A function like
```
__itt_timestamp __itt_get_timestamp(void);
```
Is currently generate the tp
```
TRACEPOINT_EVENT(
  lttng_ust_itt,
  __itt_get_timestamp,
  TP_ARGS(
    void
  ),
  TP_FIELDS(
    ctf_integer(__itt_timestamp, ittResult, ittResult)
  )
)
```
This is obviously wrong. The `TP_FIELDS` are correct... So yeah, easy fix

### DIFF
--- a/utils/gen_probe_base.rb
+++ b/utils/gen_probe_base.rb
@@ -13,7 +13,7 @@ $tracepoint_lambda = lambda { |provider, c, dir = nil|
       TP_ARGS(
   EOF
   print '    '
-  if (c.parameters.nil? || c.parameters.empty?) && !(c.has_return_type? && dir == :stop)
+  if (c.parameters.nil? || c.parameters.empty?) && !(c.has_return_type? && dir != :start)
     print 'void'
   else
     params = []


TApplencourt:
Look like `ITT_collector.h` is implementing this function `https://github.com/intel/pti-gpu/blob/0f5841470aec005b5ba461fc7f31cc9b059b5d5e/tools/unitrace/src/itt/itt_collector.h#L793`, where my current 2h fresh understanding is that they should not.  So yeah more digging  / understanding to do for ITT and future tomorrow me, to continue clarify this thing.

The ITT header are here https://github.com/intel/ittapi/blob/master/src/ittnotify/ittnotify_static.c

TApplencourt:
Oh and by the way `__itt_string_handle_create`  is another callback who return something. Yeah, ITT is weird. They forward the implementation to the callbacks...

Kerilk (approved):

PR #430, opened 2025-10-13
H2yaml by default

TApplencourt:
Just so I have a PR, so I know if H2yaml break stuff (easier that to check actions tabs...)

### DIFF
--- a/.github/actions/setup-thapi/action.yml
+++ b/.github/actions/setup-thapi/action.yml
@@ -103,7 +103,7 @@ runs:
 
     - name: Install H2yaml
       shell: bash
-      run: pip install git+https://github.com/TApplencourt/h2yaml # @v0.2.1 # Todo, put it back before merging
+      run: pip install git+https://github.com/TApplencourt/h2yaml@v0.3.0
 
     - name: Set Efficios environment variables
       shell: bash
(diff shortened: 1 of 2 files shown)


TApplencourt:
Will continue tomorow. Maybe due to the macro handled. As expcet Cast
```
- name: _ze_init_flag_t
  members:
  - name: ZE_INIT_FLAG_GPU_ONLY
    val: ZE_BIT(0)
  - name: ZE_INIT_FLAG_VPU_ONLY
    val: ZE_BIT(1)
  - name: ZE_INIT_FLAG_FORCE_UINT32
    val: '0x7fffffff'
```
h2yaml:
```
- members:
  - name: ZE_INIT_FLAG_GPU_ONLY
    val: 1
  - name: ZE_INIT_FLAG_VPU_ONLY
    val: 2
  - name: ZE_INIT_FLAG_FORCE_UINT32
    val: '0x7fffffff'
  name: _ze_init_flag_t
```

Maybe that trip down the metababel yaml gen

Kerilk:
For `unsigned long int` to `unsigned long` error in the CI, this would need to be added on the compatibility side. I don't have the information in `cast-to-yaml` as to how the type was written, so I defaulted to always output the optional int part.

TApplencourt:
I'm doing it already (https://github.com/TApplencourt/h2yaml/blob/af80fba27af117e9476b29d9b7ac470881001b68/h2yaml.py#L433). In fact i was always doing it : ) It's the first compact I did : )

So I think it's more a typedef problem or something... Didn't have time to have a look today : (

Kerilk:
I assume next step will remove most of the patches and comparison with cast-to-yaml.

PR #438, opened 2025-10-20
ITT integration tests

nscottnichols:
Initial integration tests for ITT.

### DIFF
--- a/integration_tests/backend_itt.bats
+++ b/integration_tests/backend_itt.bats
@@ -1,39 +1,21 @@
 bats_require_minimum_version 1.5.0
 
-setup() {
-  # Useful paths
-  ITT_TEST_ROOT="${BATS_TEST_DIRNAME}"
-  ITT_SRC_DIR="${ITT_TEST_ROOT}"
-  ITT_TMP_DIR="${ITT_TEST_ROOT}/tmp"
-  mkdir -p "${ITT_TMP_DIR}"
-}
-
-teardown_file() {
-  # Clean up any traces we created in this file
-  rm -rf "${ITT_TMP_DIR}"
-}
-
 @test "ITT (C): trace contains __itt_task_begin events" {
-  gcc ${ITT_SRC_DIR}/itt_example.c -I${ITTAPI_ROOT}/include ${ITTAPI_ROOT}/lib/libittnotify.a -o ${ITT_TMP_DIR}/itt_example
-  local out_file="${ITT_TMP_DIR}/itt_out_c.txt"
-  $IPROF --backends itt --analysis-output ${out_file} -- ${ITT_TMP_DIR}/itt_example
-  grep "Example.Domain:Task 2" ${out_file}
-  grep "Example.Domain:Task 1" ${out_file}
+  gcc ./integration_tests/itt_example.c -I${ITTAPI_ROOT}/include ${ITTAPI_ROOT}/lib/libittnotify.a -o itt_example
+  $IPROF --backends itt --analysis-output ./itt_out_c.txt -- ./itt_example
+  grep "Example.Domain:Task 2" ./itt_out_c.txt
+  grep "Example.Domain:Task 1" ./itt_out_c.txt
 }
 
 @test "ITT (Python, context manager): trace contains ITT task events" {
-  local script=${ITT_SRC_DIR}/itt_example_context-manager.py
-  local out_file="${ITT_TMP_DIR}/itt_out_py_context-manager.txt"
-  $IPROF --backends itt --analysis-output ${out_file} -- python3 ${script}
-  grep "Example.Domain:Task 2" ${out_file}
-  grep "Example.Domain:Task 1" ${out_file}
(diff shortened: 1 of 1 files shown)


integration_tests/backends/itt/itt.bats
```diff
@@ -0,0 +1,161 @@
...
+    return 1
+  fi
+
+  (
+    cd "${ITT_SRC_DIR}/itt_example_c" || exit 1
+    "${cc}" itt_example.c -O2 -I"${inc}" -L"${libdir}" -littnotify -o itt_example >/dev/null 2>&1
+  )
+  [[ -x "${exe}" ]] && printf "%s" "${exe}" && return 0
+  return 1
+}
```
TApplencourt: We know what we build, can we just do 
```
 "${cc}" itt_example.c -littnotify -o itt_example 
```
The INCLUDE path an LIBRARY_PATH should have been set in the `presubmit.yml` during configuration

integration_tests/backends/itt/itt.bats
```diff
@@ -0,0 +1,161 @@
...
+
+  run "${IPROF:-iprof}" --no-analysis --backends itt --trace-output "${trace_dir}" -- "${c_exe}"
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run _bt_pretty "${trace_dir}"
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [ "$(echo "$output" | grep -c 'lttng_ust_itt:__itt_task_begin')" -ge 2 ]
+  echo "$output" | grep -q 'Task 1'
```
TApplencourt: We can check how we do in other backen (I think @thilinarmtb  did add some test like this). 
So at least we share the infrastructure.

But maybe if we can just skip and just do a normal run and check in the tally. Something ismple as
```

@test "default_summary" {
   total_count=$( $IPROF $THAPI_TEST_BIN | awk -F'|' '/Total/ {print int($4)}' )
   [ "$total_count" -ge 1 ]
}
```
(as if the tally have faild, then everything else have failed :P )

integration_tests/backends/itt/src/itt_example_c/itt_example.c
```diff
@@ -0,0 +1,21 @@
...
+
+int main(void) {
+    __itt_domain* domain = __itt_domain_create("Example.Domain");
+    __itt_string_handle* task1 = __itt_string_handle_create("Task 1");
+    __itt_string_handle* task2 = __itt_string_handle_create("Task 2");
+
+    printf("Starting ITT example...\n");
+
+    __itt_task_begin(domain, __itt_null, __itt_null, task1);
+    for (volatile int i = 0; i < 100000000; ++i);
```
TApplencourt: same as in python, maybe no need any loop : )

integration_tests/backends/itt/itt.bats
```diff
@@ -0,0 +1,161 @@
...
+  else
+    local bbt_bin="${BBT:-babeltrace_thapi}"
+    if command -v "${bbt_bin}" >/dev/null 2>&1; then
+      "${bbt_bin}" --no-restrict "$@"
+    else
+      echo "SKIP: Neither 'babeltrace2' nor '${bbt_bin}' available." >&2
+      return 127
+    fi
+  fi
+}
```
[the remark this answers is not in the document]
nscottnichols: This is what we do for the CXI sampling though (I guess because there is no tally). Would you rather I grep the output of `iprof`?

integration_tests/backends/itt/itt.bats
```diff
@@ -0,0 +1,161 @@
...
+except Exception:
+    sys.exit(2)
+sys.exit(0)
+PY
+  case $? in
+    0) return 0 ;;
+    2) skip "Python package 'ittapi' not installed; skipping Python ITT tests." ;;
+    *) skip "Unable to check Python 'ittapi'; skipping Python ITT tests." ;;
+  esac
+}
```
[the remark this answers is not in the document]
nscottnichols: I was wondering what you would think of that. Will remove.

integration_tests/backends/itt/itt.bats
nscottnichols: Will remove.

integration_tests/backends/itt/itt.bats
```diff
@@ -0,0 +1,161 @@
...
+  [ "$status" -eq 0 ]
+
+  run _bt_pretty "${trace_dir}"
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [ "$(echo "$output" | grep -c 'lttng_ust_itt:__itt_task_begin')" -ge 2 ]
+  echo "$output" | grep -q 'Task 1'
+}
+
+@test "ITT (Python, context manager): trace contains __itt_task_begin events" {
```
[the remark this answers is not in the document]
nscottnichols: Will change.

integration_tests/backends/itt/src/itt_example_c/itt_example.c
nscottnichols: Will remove loop.

PR #445, opened 2025-10-22
[ITT] initial ITT make check tests

nscottnichols:
Add initial `make check` tests for ITT backend.


TApplencourt:
You just need to rebase : )

PR #448, opened 2025-10-29
Perfetto Refactoring

TApplencourt:
This PR refactor the timeline. We have two main goal:
- Easy creation of arbitrarily nested tracks.
- Fast Lookup of those tracks

The code is structured as follow:
- The `trace` class is the only user exposed class. It provide an `add_event_slice`  and `add_event_counter` function.
Those function API look like
```
  template <typename... KeyArgs>
  void add_event_slice(const std::string &name,
                       uint64_t begin,
                       uint64_t end,
                       std::function<std::vector<std::string>(void)> get_names,
                       const std::tuple<KeyArgs...> &caching_key);
```
For example

```
  dispatch->trace->add_event_slice(
      name, ts, ts + dur,
      [=] {
        return std::vector{"Hostname " + std::string{hostname}, "Process " + std::to_string(vpid),
                           "Thread " + std::to_string(vtid)};
      },
      std::make_tuple(hostname, vpid, vtid));
}
```
This will create recursively tracks `Hostname / Process /Threads` and then add the `name` event into this.
If the key is matching, the string creation of the "path" (and recursive lookup to get the associated track) will be avoided.  This put us on part performance wise to the old implementation.

Note on ownership: We use smart pointer everywhere. The exception is the `raw` pointer that each Track have to the Trace. (This is due to the fact, that it's painfull to create a `shared_pointer` from `this`).  This pointer is used to increment the global track `uuid` and have access to the global `string interning`. 

Note on caching: It's obviously user responsibility to use unique key for the their track.  For now multiple intance of Track shared the same cache. If we required, we can improve user-exepeirence in this futur by adding some `trace instance` uuid to the caching mechanisme, but it's not required for now so I keep it clean.

### DIFF
--- a/xprof/btx_timeline.cpp
+++ b/xprof/btx_timeline.cpp
@@ -14,7 +14,7 @@
 
 #include "perfetto_pruned.pb.h"
 
-enum { TRUSTED_PACKED_SEQUENCE_ID = 88,
+enum { TRUSTED_PACKET_SEQUENCE_ID = 88,
        MAX_EVENT_PER_TRACE_CHUNK = 100'000 };
 
 // Forward Declare.
@@ -48,9 +48,9 @@ public:
       return std::nullopt;
 
     // Sequence Flags:
-    // The packet calling `get_interned_id` will used interned string
+    // The packet calling `get_interned_id` will use an interned string
     int flags = perfetto_pruned::TracePacket::SEQ_NEEDS_INCREMENTAL_STATE;
-    // If it the first packet to use interned string we need to clean the state
+    // If it is the first packet to use interned string, we need to clean the state
     if (name_to_iid_.empty()) {
       flags = flags | perfetto_pruned::TracePacket::SEQ_INCREMENTAL_STATE_CLEARED;
     }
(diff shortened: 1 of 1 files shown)


TApplencourt:
I will put this branch in my home, and let people tests. UI is hard. For example maybe we can "open by default" some track (I think it's possible?).
We lost the color for the merger track, but can be added back (maybe tricky 🤷🏽 )

Lesson learned during the refactoring. "It's just a tree"

TApplencourt:
Ok ready to review

xprof/perfetto_prunned.proto
```diff
@@ -69,37 +39,19 @@ message TrackEvent {
...
+  oneof correlation_id_field {
+    uint64 correlation_id = 52;
+  }
 }
 
 message TracePacket {
   optional uint64 timestamp = 8;
 
   enum SequenceFlags {
-    SEQ_UNSPECIFIED = 0;
```
Kerilk: Any reason to remove this value?

TApplencourt:
Ok, Thanks to @Kerilk review I think i'm done.

Let me try to explain the code in the description of this PR

TApplencourt:
Description updated

nscottnichols (approved):
LGTM!

PR #449, opened 2025-10-31
Remove template parameter from `strip_event_class_name`

thilinarmtb:
Passing `strlen()` runtime value to the template parameter is not valid C++. This PR
fixes it. Without the fix, we get a compiler error with `clang` (but most of the other
compilers are okay with it).

PS: It may be a few lines, but there was a lot of sweat, tears and blood behind this 😃

### DIFF
--- a/utils/xprof_utils.hpp
+++ b/utils/xprof_utils.hpp
@@ -201,15 +201,15 @@ private:
 };
 
 // ClassName Striping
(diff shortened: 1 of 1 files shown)


TApplencourt (approved):

TApplencourt:
I let you merge when the CI is done : ) 
Thanks a lot! (and thanks @nliber  too : ) )

PR #452, opened 2025-11-05
[ITT] pretty print

nscottnichols:
Allows for pretty print to work with the ITT backend.

### DIFF
--- a/backends/itt/gen_itt_library_base.rb
+++ b/backends/itt/gen_itt_library_base.rb
@@ -35,12 +35,6 @@ def to_snake_case(str)
   str.gsub(/([A-Z][A-Z0-9]*)/, '_\1').downcase
 end
 
-def to_class_name(name)
-  mod = to_name_space(name)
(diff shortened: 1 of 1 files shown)


backends/itt/gen_itt_library.rb
```diff
@@ -0,0 +1,337 @@
...
+
+  pair = ->(sym_val) { sym, val = sym_val; "#{sym.inspect}, #{val}" }
+
+  puts <<~RUBY
+    #{to_class_name(name)} = ittenum #{to_ffi_name(name)},
+      [ #{resolved.map(&pair).join(",\n        ")} ]
+  RUBY
+end
+
+def print_itt_object(object)
```
nscottnichols: `print_<be>_object` does the same thing as `print_pointer_type`. Why have a specialization here instead of removing the code and doing `close_type` in the pointer part of the big if...else section (line 321)?

nscottnichols: Cool, I figured there was some history behind it. I'll leave it as is for now to mirror the other backends (and be cleaned up in a future refactor).

PR #453, opened 2025-11-12
Refactoring of pretty printer.

Kerilk:
Cleanup and refactoring.
Added generated file comparison for itt that was missed.


Kerilk:
OK, most probably won't be able to work more on it be fore a couple of weeks, and it feels a bit too useful to let rot, so I think we should merge, unless we find some glaring issues during the review process.
 * Refactoring of pretty printer generators
 * Uniformization of Makefiles
 * Fix ITT ci

TApplencourt:
>+422-1,267

Amazing! 
(Now back to prepare by bag for SC...)

PR #454, opened 2025-11-17
Add `thapi_start()` and `thapi_stop()`

thilinarmtb:
I created `lttng_ust_toggle` event category and implemented `thapi_start()` and `thapi_stop()`
user APIs (in a similar spirit to `cudaProfilerStart()` etc.). The library which the user has to link
against is named `libThapi.so`.

Usage of the API is as follows:
```C
#include <thapi.h>
//
// code
//
thapi_start();
// code to be profiled
thapi_stop();
//
// more code
//
```

If `libThapi.so` is linked, `thapi_auto_stop()` (which is not part of the user API) is called automatically
during the library load time using `__attribute__((constructor))`. I added a check in `configure.ac` to
check if the compiler supports `__attribute__((constructor))`.

If the user runs `iprof` with `--toggle-on`, the traces generated by `thapi_auto_stop()`, `thapi_start()`
and `thapi_stop()` will be respected by `babeltrace` (otherwise there is no change to `iprof/babeltrace`
behavior). Since `thapi_auto_stop()` is called automatically, `babeltrace` will filter out all of the traces
till it see a `thapi_start()`. Trace filtering out will start again once the user call `thapi_stop()`. This API
only affects the trace collection of the calling process/thread.

**P.S**: This feature requires `lttng-ust >= 2.12.8`. Specifically, this [fix](https://github.com/lttng/lttng-ust/commit/a8fafb675a9f580f6a889223e26664ea11cb0c99).

### TODO
- [x] ~Respect ~`lttng_ust_profiler_*`~ `lttng_ust_toggle_*` events when calculating statistics~
- [x] Fix failing checks
- [x] Add the `filter.toggle.toggle` to `babeltrace` graph


TApplencourt:
@thapi-local-reviewer review

TApplencourt:
sorry @thilinarmtb  just playing with some bot :p

PR #456, opened 2025-12-01
Fix (some) integration test failures on Aurora

thilinarmtb:
#### Failing Tests
- `default_summary` test fails on Aurora since the current `awk` command prints two numbers
    - One from `ITT` and the other from `OpenCL`
    - Passed the `--backend cl` to only run the `OpenCL` backend
-   `no-analysis_output` test fails as `iprof` merges `stderr` and `stdout`
    - There are `Missing symbol ...` warnings on Aurora when `iprof` tries to load `libOpenCL.so` in `stdout`
    - This PR separates `stdout` and `stderr`.

#### Steps to Reproduce
- On a compute node on Aurora do the following:
    - `module restore`
    - `module load oneapi mpich`
- After setting up spack, do:
    - `spack build-env thapi@develop+archive+dev bash` (needs https://github.com/argonne-lcf/THAPI-spack/pull/98)
    - Build and install thapi
- Run the tests by doing for example:
    - `THAPI_BIN_DIR=<thapi/install/bin> bats ./integration_tests/general.bats`

### DIFF
--- a/integration_tests/general.bats
+++ b/integration_tests/general.bats
@@ -47,6 +47,11 @@ teardown_file() {
    [[ "$err2" =~ "THAPI: Trace location" ]]
 }
 
+@test "stderr_output" {
+   run --separate-stderr $IPROF -- bash -c "echo \"error\" >&2"
+   [[ "$stderr" =~ "error" ]]
+}
+
 @test "no-analysis_all" {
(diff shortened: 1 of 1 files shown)


thilinarmtb:
Done! Thank you!

TApplencourt (approved):

TApplencourt:
Thanks!

PR #459, opened 2025-12-12
Check for test tools only in `$PATH`

thilinarmtb:
We check if the user variable path exists in `setup_suite.sh`, not if the plain command
exists in`PATH` variable. Without this change, `bats` can't find test tools in PATH once
I do `spack load thapi@develop+test-dependencies`. See [here](https://gitlab-ci.alcf.anl.gov/thilina/THAPI-spack-0/-/jobs/164232) for an example.


TApplencourt:
let just use `iprof` everywhere :)

TApplencourt:
>You have an error in your yaml syntax on line 136

lol sorry!

PR #462, opened 2025-12-17
Add `shfmt` workflow

thilinarmtb:


TApplencourt:
`lstat -w: no such file or directory` :p

thilinarmtb:
> `lstat -w: no such file or directory` :p

I don't even use the `-w` option so not sure what is going on. I am going to
give it some time and try again later to see if time will resolve this problem 😄

PR #464, opened 2026-01-29
Add a THAPI build with Fil-C

thilinarmtb:
C can be memory safe too!  

Website: https://fil-c.org/


TApplencourt:
memory safe but cannot run :p

PR #472, opened 2026-02-24
Add rubocopt Lint (and run it)

TApplencourt:
Add  (and run)
- clang-format
- yaml-fmt
- rubocopt

Those will be linted, see https://github.com/argonne-lcf/THAPI/blob/85621c7d5407d292fbc0fdef2bcee383e780a476/.github/workflows/lint.yml

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -90,6 +90,8 @@ jobs:
         files: |
           **.cpp
           **.hpp
+          **.c
+          **.h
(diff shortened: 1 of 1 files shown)


TApplencourt:
Ok Finally good. Should help maintain some consitancy. The failure are due to `include.c` who changed, nothing else.
We can fine-tune our styles (clang-format, rubocopt and yalmft) as we go.

PR #485, opened 2026-03-30
Continue H2yaml migration

TApplencourt:
First one a set of PR,
We just remove `extract.rb`.
In next PR, we will remove other workarround who was required in by the old approach.

But this may change the `api_*.yaml`, so better to do it in too step.


.github/actions/setup-thapi/action.yml
```diff
@@ -34,7 +34,7 @@ runs:
       packages: gcc g++ ruby ruby-dev elfutils libelf-dev libpopt-dev libdw-dev libprotobuf-dev protobuf-compiler valgrind libc6-dbg libglib2.0-dev libnuma-dev
   - name: Install Ruby Dependencies
     shell: bash
-    run: sudo gem install cast-to-yaml nokogiri babeltrace2 opencl_ruby_ffi metababel
+    run: sudo gem install nokogiri babeltrace2 opencl_ruby_ffi metababel cast-to-yaml
```
TApplencourt, line 37: `cast-to-yaml ` is not required anymore for `devel`, but still on this PR to do the `compare step`. I will remove it in another PR after this one is done.

backends/hip/extract/hip_api.h
```diff
@@ -0,0 +1,17 @@
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
+
+#define __HIP_PLATFORM_AMD__
+#define THAPI_NO_INCLUDE
+
+#include <hip/hip_runtime_api.h>
+
```
TApplencourt, line 5: Block are here to avoid `clang-format` reordering them. Reordering them, break the `diff` for the `tracer.*` so I don't want to be confused in this PR

nscottnichols (approved):
LGTM!

PR #486, opened 2026-03-31
Continue h2yaml migration

TApplencourt:
Cuda have no more patches. \o/

The `yaml` change because of that (we support anonymous enum). We modifyed the `utils/gen_library_base.rb` to support those.


TApplencourt:
Hum `hip` have changed too, thanks CI : ). Let me fix this

TApplencourt:
All good now. Only the `yaml` change, but no change anywhere else.

PR #493, opened 2026-04-23
Fix dlopen

TApplencourt:
I was so mad about slow FS on aurora, we by pass ldd and `dlopen`.

And fix Thilina bug about rpath in some case. And give access to a binary if user what to find the lib, so they can set the path of the library to dlopen by our tracer manualy.


thilinarmtb (approved):

TApplencourt:
If it fail, we will fall back to `dlinfo`. But I we will see in pratice I suppose.

PR #510, opened 2026-06-22
Always report stderror when failing

TApplencourt:


TApplencourt:
Welldone. Super nice PR!

PR #511, opened 2026-06-22
Better support of pre-2.7 ruby

TApplencourt:
iprof (xprof.rb) uses modern ruby syntax (endless ranges), so an old ruby fails to *parse* the whole file before the version check can run, giving a raw SyntaxError instead of a friendly message.

Split iprof into a tiny old-ruby-safe wrapper (xprof_wrapper.rb.in -> bin/iprof) that checks the ruby version, then requires the implementation (xprof.rb, now in DATADIR). The wrapper uses only old, plain syntax, so the version gate runs and prints a friendly message on old ruby.

Also add a configure-time ruby >= 2.7 check (single-sourced via RUBY_MIN_VERSION) and a bats regression test.


TApplencourt:
Amazing PR Thomas!

PR #513, opened 2026-06-23
Make iprof a transparent stdio wrapper

TApplencourt:
iprof ran the user binary via IO.popen, which connects the child's stdout to a pipe. The child's libc then sees a non-tty and switches to full block-buffering, so its output only appeared once it exited (e.g. all of `iprof -m full ./a.out` arrived in one block instead of streaming).

Let the child inherit iprof's own stdout/stderr directly instead, so it sees exactly the descriptors iprof was given and picks the same buffering it would without iprof: line-buffered (streaming) on a terminal, block-buffered on a pipe/file. This also keeps stdout and stderr on their own streams.

Add an integration test driving iprof under a PTY (via `script`) to check streaming with a real libc-buffered binary, and one checking stdout/stderr stay separate. The previous test used `bash echo`, which is not libc-buffered and so never exercised the bug.

### DIFF
--- a/integration_tests/buffering_helper.c
+++ b/integration_tests/buffering_helper.c
@@ -1,0 +1,20 @@
+/*
+ * Helper for the iprof stdout/stderr streaming tests.
+ *
+ * It uses plain stdio (printf/fprintf) with NO explicit fflush, so the C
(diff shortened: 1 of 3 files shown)


TApplencourt:
Superb fix.

PR #514, opened 2026-06-24
sync_daemon_mpi: give daemon a distinct PMIX_NAMESPACE

TApplencourt:
The MPI sync daemon shares the traced application's inherited PMIx identity (PMIX_NAMESPACE). Two MPI instances claiming the same (namespace, rank) collide in the node-local SHM bootstrap and deadlock (n=128 --ppn 64) or abort with a truncation error (n=4).

Append a fixed suffix to PMIX_NAMESPACE before any MPI call so the daemon's session gets its own WORLD. Verified on 2 nodes with THAPI_SYNC_DAEMON=mpi: fix -> exit 0, full tally; control (fix disabled) -> hang. Default remains fs pending MPICH feedback.

### DIFF
--- a/xprof/sync_daemon_mpi.cpp
+++ b/xprof/sync_daemon_mpi.cpp
@@ -2,6 +2,7 @@
 #include "mpi.h"
 #include <cstdlib>
 #include <cstring>
+#include <string>
 
 using namespace daemon_proto;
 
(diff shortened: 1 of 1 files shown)


TApplencourt:
Related to https://github.com/pmodels/mpich/issues/7852

nscottnichols (approved):
LGTM!

PR #530, opened 2026-08-11
mpi: load the meta-parameters and reconcile them with the API

TApplencourt:
mpi_meta_parameters.yaml has never been loaded. Its top-level key is the Ruby symbol `:meta_parameters:`, but mpi_model.rb reads the string key `meta_parameters`, so `fetch` silently returned the default and MPI registered 0 of its 908 rows. Dead since 9ae9297 ("Backends folder").

Flipping the key alone crashes the generator: the data was machine generated as one InScalar/OutScalar row per pointer parameter and, never having been executed, drifted out of sync with the API. Reconciled by introspecting every row against mpi_api.yaml:

  - drop 246 rows on void* / const void* buffers (MPI_Allgather/sendbuf, MPI_Accumulate/origin_addr, ...). These have no lttng_type -- the payload is untyped bytes whose length lives in a separate count+datatype pair, so there is nothing to dereference. The buffer pointer itself is still traced.
  - drop 166 rows on array parameters (const int[], MPI_Request[], ...). ScalarMetaParameter requires a pointer; these need InArray/OutArray with a size argument, which the 2-element rows do not carry. Sizing them is per-function semantics (MPI_Allgatherv/recvcounts is sized by the communicator, MPI_Alltoallv's counts by nothing in the signature), not something to guess.
  - drop 6 array_of_statuses rows. They pass the pointer test but a scalar row would dump only element 0 of the array.
  - retype 58 char* rows to InString/OutString. As scalars they emitted ctf_integer(char, x_val, *x) -- a single byte of a NUL-terminated string. Direction now follows constness, which corrects three rows the generator had marked InScalar on non-const out-parameters (MPI_Get_library_version/version, MPI_Get_processor_name/name, MPI_Session_get_nth_pset/pset_name).

490 rows now register and emit: 212 ctf_integer_hex, 167 ctf_integer, 58 ctf_string, 53 ctf_sequence_text. All 536 function keys are kept so the file still lists the full API surface. The dropped rows are the ones that cannot be expressed as scalars; adding proper InArray/OutArray coverage for the buffer and array parameters is separate work.

### DIFF
--- a/backends/mpi/mpi_meta_parameters.yaml
+++ b/backends/mpi/mpi_meta_parameters.yaml
@@ -1,9 +1,7 @@
 ---
-:meta_parameters:
+meta_parameters:
   MPI_Abort: []
-  MPI_Accumulate:
-  - - InScalar
-    - origin_addr
+  MPI_Accumulate: []
   MPI_Add_error_class:
   - - OutScalar
     - errorclass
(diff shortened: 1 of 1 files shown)


Kerilk:
The backends commit didn't touch the file the blame only points to it because the file was moved. Most probably predates that.

TApplencourt:
Yeah, it orginated for the introduction of the MPI backend. Was a typo on my side :( The only thing it add realy is the `mpi_get_error_string` and `mpi_get_info`.  

I caugh it, because in the other PR for the refactoring I print a warning if no metadata are found, are the MPI backend trigered the warning. "Oh no..."

PR #540, opened 2026-09-14
Pytorch backend

DonAurelio:
Adds a pytorch backend to THAPI that traces PyTorch RecordFunction (forward + backward) ops. Enable it explicitly:

```
iprof --backends pytorch --no-analysis -- python model.py
```

**Pending follow-up work (not in this PR):**

- **Default backend.** Currently must be explicitly requested with `--backends pytorch`, and requires `--no-analysis` since there's no analysis-side decoder for this backend yet. Making it a default backend depends on adding that.
- **Second calling convention via environment variables.** `iprof -- python3 model.py` (active torch env) works; `iprof -- venv/bin/python3 module.py` (non-active env) doesn't yet — needs env vars to let the target interpreter be specified explicitly.
- **ABI version validation.** The tracer hand-copies a private PyTorch struct layout for the version it was built against; a version mismatch would silently corrupt tracing rather than fail loudly. Planned fix is a version allow-list checked at load time.

## Build: suppress two clang-only warnings in the pytorch tracer

`tracer_pytorch.cpp` is the only backend translation unit that calls LTTng's `tracepoint()` macro from C++ instead of C (every other backend does so from a `.c` file). That exposes two warnings that only clang raises, both false positives.

**Before:**
```
libTracerPytorch_la_CXXFLAGS = -std=c++17 -Wall -Wextra -Wno-unused-parameter $(WERROR) $(LTTNG_UST_CFLAGS)
```

**After:**
```
libTracerPytorch_la_CXXFLAGS = -std=c++17 -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wno-unused-private-field $(WERROR) $(LTTNG_UST_CFLAGS)
```

- `-Wno-missing-field-initializers` — LTTng's `tracepoint.h` initializes some of its internal structs with a designated initializer that only names one field (e.g. `.struct_size = ...`), leaving the rest to their default zero value — a standard, safe C idiom. C accepts this silently; C++ (under clang) does not, and treats it as a missing-initializer warning. Every other backend includes this same header from a `.c` file compiled as C, so they never trigger it; `tracer_pytorch.cpp` is the only backend source that pulls it into a C++ translation unit.
- `-Wno-unused-private-field` — `RecordFunctionCallback` in our hand-written `record_function.h` declares private fields (`start_`, `end_`, `sampling_prob_`, etc.) that our own code never reads. That's intentional: this struct is passed by value into PyTorch's real `addGlobalCallback`, so its memory layout must match PyTorch's real class byte-for-byte — the fields are read by libtorch on the other side of that call, not by us. Clang can't see that and flags them as dead; gcc doesn't have this check at all.


backends/pytorch/Makefile.am
```diff
@@ -0,0 +1,86 @@
...
+libTracerPytorch_la_CPPFLAGS = -I$(top_srcdir)/utils -I$(top_srcdir)/utils/include -I$(srcdir)/include -I./
+# -Wno-missing-field-initializers: LTTng's own tracepoint.h (not our code)
+#   aggregate-initializes structs without naming every field; clang is
+#   stricter than gcc about this under -Wextra.
+# -Wno-unused-private-field: RecordFunctionCallback's fields exist only to
+#   match PyTorch's real class layout byte-for-byte (see the comment in
+#   include/ATen/record_function.h) -- the real libtorch reads them, we
+#   never do, by design.
+libTracerPytorch_la_CXXFLAGS = -std=c++17 -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wno-unused-private-field $(WERROR) $(LTTNG_UST_CFLAGS)
+libTracerPytorch_la_LDFLAGS = $(LTTNG_UST_LIBS) -Ldummy_libs -ltorch_cpu -avoid-version -module
```
TApplencourt, line 84: why libtorch_cpu and not just `libtorch`? `libtorch` 
We should verify that it work with the GPU backend of pytorch too if not done already

DonAurelio: LibTracerPytorch.so calls at::addGlobalCallback() and RecordFunction::name(), both are implemeted in libtorch_cpu.so

TApplencourt (approved):

TApplencourt:
Thanks. So next step `interval`, `tally` and `timeline`. 
Also potentialy the pretty print, so that `iprof --backends pytorch --trace -- python model.py`  work

